PR #11, opened 2025-09-16
Library docs improvements

rlycx:
a couple of things to make the generated docs nicer, at least for the producer module

### DIFF
--- a/libs/opsqueue_python/src/common.rs
+++ b/libs/opsqueue_python/src/common.rs
@@ -232,7 +232,7 @@ impl Chunk {
 #[pymethods]
 impl Chunk {
     fn __repr__(&self) -> String {
-        format!("{:?}", self)
+        format!("{self:?}")
     }
 }
 
@@ -261,14 +261,14 @@ impl ChunkFailed {
 #[pymethods]
 impl ChunkFailed {
     fn __repr__(&self) -> String {
-        format!("{:?}", self)
+        format!("{self:?}")
     }
 }
 
 #[pymethods]
 impl Strategy {
     fn __repr__(&self) -> String {
-        format!("{:?}", self)
+        format!("{self:?}")
     }
 }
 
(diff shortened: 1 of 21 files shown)


libs/opsqueue_python/pyproject.toml
```diff
@@ -60,7 +60,3 @@ test = [
 # We ensure tests never rely on global state,
 # by running them in a random order, and in parallel:
 addopts = "--random-order --workers=auto"
-# Individual tests should be very fast. They should never take multiple seconds
-# If after 20sec (accomodating for a toaster-like PC) there is no progress,
-# assume a deadlock
-timeout=20
```
Qqwy, line 66: I guess the CI servers are sometimes a 'toaster-like PC'? :see_no_evil:
Are you sure you want to remove the deadlock protection? Maybe bumping this to, say, 60 seconds, would be enough?

libs/opsqueue_python/src/common.rs
```diff
@@ -232,7 +232,7 @@ impl Chunk {
 #[pymethods]
 impl Chunk {
     fn __repr__(&self) -> String {
-        format!("{:?}", self)
+        format!("{self:?}")
```
Qqwy, line 235: Thank you :+1:

Qqwy (approved):
Thank you for the nice improvements in the documentation, and better formatting <3

libs/opsqueue_python/pyproject.toml
Qqwy: The plot thickens: Why does `pytest-timeout` not get loaded as part of the Nix overlay?

rlycx: it's not listed in default.nix like for example `pytest-parallel`. adding it there also seems to fix the issue

rlycx:
@OpsBotPrime merge

PR #17, opened 2025-10-08
Allow http(s) scheme in the producer client host string

### LINKED ISSUE #13: Don't append `http` if it's already present
Ran into a funny issue:

If you try passing the following url to the Producer `http://example.com` it will try to call the following endoint `http://http://example.com`.

If possible, if would be nice not to append the `http://` to the url if it's already present.

rlycx:
fixes #13


rlycx:
@OpsBotPrime merge

OpsBotPrime:
Rebased as 47904d2c15742d5f6d1197571ac30775613c2448, waiting for CI …

PR #19, opened 2025-10-17
Update cargo dependencies

rlycx:
It has been a while, and i noticed we still have an open PR from dependabot to upgrade deps, so hereby some love for our Cargo.lock. 

Closes #10.


Qqwy (approved):
There's very little to actually review for me here. Since the tests still pass, this should be fine 👍

PR #49, opened 2026-01-29
Pin remaining GitHub Actions

tijmendj:
Part of https://github.com/channable/devops/issues/13319.

Some people argued that it would be more consistent if we also pinned GitHub owned (i.e. github.com/actions/checkout) Actions.


Qqwy:
@OpsBotPrime retry on Friday

PR #51, opened 2026-01-30
In tests on CI, store the test DBs in memory to fix self-hosted Semaphore runners sometimes hanging

Qqwy:
This is with high certainty the cause of the Semaphore hangs we have seen in the past, which took @rlycx, @ReinierMaas, @SanderHageman and myself a huuuge amount of work to debug: There is an interaction between writing a file to `/tmp/` (which, to my surprise, is _not_ a `tmpfs` in Semaphore) on Butterfs when using many small fsyncs (which as it turns out, Sqlite does when initializing a new DB, esp in WAL mode), and then doing this for dozens of tests in parallel.

This will cause one of these `fsync`s to hang for many minutes on end, with the OS busy doing some kind of disk vacuuming.

This PR changes Opsqueue to run in a 100% in-memory mode in the integration test suite instead.

### DIFF
--- a/libs/opsqueue_python/tests/conftest.py
+++ b/libs/opsqueue_python/tests/conftest.py
@@ -4,7 +4,6 @@ from contextlib import contextmanager, ExitStack
 from typing import Generator, Callable, Any, Iterable
 import multiprocessing
 import subprocess
-import uuid
 import os
 import pytest
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


OpsBotPrime:
Your merge request has been denied, because merging on Fridays is not recommended. To override this behaviour use the command `merge on Friday`.

Qqwy:
@OpsBotPrime merge on Friday

OpsBotPrime:
Rebased as 1f4f66b22cdbd34fb2fece735fbce0ad507084e4, waiting for CI …

PR #53, opened 2026-02-03
Update most Rust dependencies

Qqwy:
This PR updates most of Opsqueue's Rust dependencies.

The one exception is the `tracing` and `opentelemetry` set of libraries, as there are some big API changes in there. So that will be done in a separate PR.


Qqwy:
@OpsBotPrime merge

PR #55, opened 2026-02-04
Update the Rust OTel Tracing crates, moving to their new API

Qqwy:
This was a bit tricky, because turns out we were relying on a bit of a loose approach where we would set the parent of some spans after they were constructed:

* Make sure the trace context is still propagated properly accomodating for the changes introduced in https://github.com/tokio-rs/tracing-opentelemetry/pull/202 that no longer allowed setting the parent of an already-built or alread-entered span.
* Remove unused OTel-related dependency

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -137,28 +137,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-stream"
-version = "0.3.6"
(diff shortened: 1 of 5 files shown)


Qqwy:
@OpsBotPrime merge

PR #60, opened 2026-02-06
Makes passing the 'rustToolchain' optional when using pkgs.callPackage to build opsqueue.nix or opsqueue_python.nix

Qqwy:
Makes passing the 'rustToolchain' optional when using pkgs.callPackage to build opsqueue.nix or opsqueue_python.nix.

They can still be overridden by specifying a different Rust version in the overlay. But if kept at its default, will use the version from the `rust-toolchain.toml` in the opsqueue repo itself.

By making this optional, users of the library do not need to care about which Rust version they are using. Except that if they _are_ using a particular version for other parts of their software, they can ensure they use the same (to prevent building deps twice, for example).

### DIFF
--- a/libs/opsqueue_python/opsqueue_python.nix
+++ b/libs/opsqueue_python/opsqueue_python.nix
@@ -3,8 +3,14 @@
   pkgs,
   lib,
 
-  # Rust version. (Override this with an overlay if you like)
-  rustToolchain,
+  # Downstream users can override which Rust version is used.
+  # But this is opt-in. By default we'll use the same version
+  # that we use in this repo for development.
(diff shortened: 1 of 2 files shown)


OpsBotPrime:
Rebased as 87d250df07b0dec52e00ba4bdb90bfbfc5c0305c, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/e07b20c6-3974-4a88-8ab1-f56fe860750c?pipeline_id=e76a4ce2-ac41-42e1-a6bb-5b1b6b495f46) started.

PR #63, opened 2026-02-06
Improved integration test suite

Qqwy:
- [x] Bumps test-only dependencies
- [x] Gets rid of using the multiprocessing `fork` strategy which was causing regular hangs of the integration test runner. This required making the things we pass around `pickle`-able, and using `multiprocess` instead of `multiprocessing` so we can pass closures around.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -10,9 +10,9 @@ let
       mypy
       uv
       pytest
-      pytest-random-order
-      pytest-parallel
+      pytest-xdist
       pytest-timeout
+      multiprocess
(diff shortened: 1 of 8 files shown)


OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/ee400d5b-473b-4a80-95d6-8ee0a3bc2d0c?pipeline_id=ac927631-edfe-4f16-ac00-0af2d5876643) started.

PR #89, opened 2026-04-01
Perform WAL checkpointing ten times per second

Qqwy:
Before: once per second
With this PR: ten times per second.

### DIFF
--- a/opsqueue/src/db/mod.rs
+++ b/opsqueue/src/db/mod.rs
@@ -256,7 +256,7 @@ impl DBPools {
     }
 
     pub async fn periodically_checkpoint_wal(&self) {
-        const EXPLICIT_WAL_CHECK_INTERVAL: Duration = Duration::from_secs(1);
+        const EXPLICIT_WAL_CHECK_INTERVAL: Duration = Duration::from_millis(100);
         let mut interval = tokio::time::interval(EXPLICIT_WAL_CHECK_INTERVAL);
         loop {
             let _ = self.perform_explicit_wal_checkpoint().await;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ members = [
 ]
 
 [workspace.package]
-version = "0.34.3"
+version = "0.34.4"
 
 
 [workspace.lints.clippy]


ReinierMaas (approved):
🚀 Let's get the WAL size under control.

Qqwy:
@OpsBotPrime merge and tag

OpsBotPrime:
Rebased as 3104526cd8f92b6f190487f0f98bb169246a43bb, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/1cffbdfe-1d74-4b32-907d-dbb14198bf82?pipeline_id=e5dc096a-010e-44fd-8fd0-ef5e88b46510) started.

opsqueue/src/db/mod.rs
```diff
@@ -256,7 +256,7 @@ impl DBPools {
     }
 
     pub async fn periodically_checkpoint_wal(&self) {
-        const EXPLICIT_WAL_CHECK_INTERVAL: Duration = Duration::from_secs(1);
+        const EXPLICIT_WAL_CHECK_INTERVAL: Duration = Duration::from_millis(100);
         let mut interval = tokio::time::interval(EXPLICIT_WAL_CHECK_INTERVAL);
```
bertptrs, line 260: I think you want to set the missed tick strategy, otherwise this will repeatedly issue checkpoints whenever the checkpoints takes too long. This will be exacerbated by the interval being shortened.

```suggestion
        let mut interval = tokio::time::interval(EXPLICIT_WAL_CHECK_INTERVAL);
        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
```

The default behaviour is "burst", i.e. fire all missed ticks right now. This is not what we want here. We don't actually rely on the number of checkpoints, we just want to fire them regularly.

bertptrs (changes requested):

OpsBotPrime:
@Qqwy I tagged your PR with [v48](https://github.com/channable/opsqueue/releases/tag/v48). Please wait for the build of 3104526cd8f92b6f190487f0f98bb169246a43bb to pass and don't forget to deploy it!

opsqueue/src/db/mod.rs
Qqwy: Doing a checkpoint when there is (almost) nothing to checkpoint is (almost) free, though it will of course add to the congestion on the one writer DB connection.

Changing to `Skip` is a nice next step. I'll also add explicit checkpoints to the two write-heavy points of the app, in a follow-up PR.

PR #96, opened 2026-04-20
Allow overriding Python version when calling the opsqueue_python Nix package

Qqwy:
Without this, if e.g. a user of the Nix Python package builds their own python that is separate from `pkgs.python3` (e.g. rather than overriding that one in an overlay, they create a separate one), it is likely that the resulting wheel / compiled FFI module is not placed in the right location, which makes the build fail (at the 'can I find all the imports' step).

This PR fixes that, by adding `python = ...`  to the parameters of `opsqueue_python.nix` meaning that a user can do `callPackage "......./opsqueue_python.nix" { python = myOtherPython; };` and be merry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,16 @@ Opsqueue's client libraries and binary itself are also available through `niv`.
 2. Package the now available `opsqueue` library as part of your overlay, using e.g.
 
 ```nix
-opsqueue = self.callPackage (sources.opsqueue + "/libs/opsqueue_python/opsqueue_python.nix") { };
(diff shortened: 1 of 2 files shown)


Qqwy:
@OpsBotPrime merge

OpsBotPrime:
Rebased as 54475ae42b6307565fe24b03fa1381acd692f841, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/88fbcdf8-c917-4e99-8bce-70b366015b0f?pipeline_id=5594510c-1c56-4263-bd7a-aebd7fdfb505) started.

PR #100, opened 2026-04-24
Include a `py.typed` marker file with the Python client

### LINKED ISSUE #8: Add `py.typed` to Python client library
The code is typed, but due to lack of a `py.typed` in the source (and final distribution) type checkers will not know this.

Would be good to add this file, ensure it is packaged and add the Trove classifier to indicate on PyPI that this package is typed.

isomorpheme:
Closes https://github.com/channable/opsqueue/issues/8. I've checked that the `py.typed` file is included when building via Nix, and then using `reveal_type` on `ProducerClient.get_submission_status` changes mypy reporting
```
note: Revealed type is "Any"
```
into
```
note: Revealed type is "def (submission_id: Any) -> Any | None"
```

So this doesn't entirely make types available, due to https://github.com/PyO3/pyo3/issues/5137, but it improves the situation.

### DIFF
--- a/libs/opsqueue_python/pyproject.toml
+++ b/libs/opsqueue_python/pyproject.toml
@@ -35,6 +35,7 @@ Issues="https://github.com/channable/opsqueue/issues"
 features = ["pyo3/extension-module"]
 python-source = "python"
 module-name = "opsqueue.opsqueue_internal"
+include = ["**/py.typed"]
 
 # We specify test-dependencies here (and have them loaded through `.setup_local_venv.sh`)
 # because otherwise `pytest` won't be able to see the locally build python package!
--- a/libs/opsqueue_python/opsqueue_python.nix
+++ b/libs/opsqueue_python/opsqueue_python.nix
@@ -44,7 +44,7 @@ let
   };
 
   # Full set of files to build the Python wheel on top
-  pythonFileFilter = path: _type: builtins.match "^.*\.(py|md)$" path != null;
+  pythonFileFilter = path: _type: builtins.match "^.*(py.typed|\.(py|md))$" path != null;
   wheelFileFilter = path: type: (pythonFileFilter path type) || (rustCrateFileFilter path type);
   wheelSrc = lib.cleanSourceWith {
     src = ../../.;


isomorpheme:
I've created https://github.com/channable/opsqueue/issues/101 as a possible followup step.

libs/opsqueue_python/pyproject.toml
```diff
@@ -35,6 +35,7 @@ Issues="https://github.com/channable/opsqueue/issues"
 features = ["pyo3/extension-module"]
 python-source = "python"
 module-name = "opsqueue.opsqueue_internal"
+include = ["**/py.typed"]
```
isomorpheme, line 38: I don't know how to check if this does what I expect it to; I don't know how to inspect the results of `just build-python`. Would appreciate pointers!

Qqwy: Does [this section of the README](https://github.com/channable/opsqueue#python-client-library) help?

Qqwy (approved):

libs/opsqueue_python/pyproject.toml
isomorpheme: It didn't per se, but I figured it out: `just build-python` builds a wheel in `/tmp` and cleans it up before finishing, so instead I used `maturin build` which outputs it locally. It turns out that `.whl` files are secretly just zip archives, so I checked and the `py.typed` is indeed included. :tada:

Qqwy:
Thank you very much for this improvement!

I'll combine it into the next release together with the other open PRs. That way our users don't need to bump multiple times in quick succession.

@OpsBotPrime merge

OpsBotPrime:
Rebased as a79b0d8fd0da43660c424c7bc47d8b5d1442be4b, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/8e99fcd1-963f-4943-a682-6559057370b5?pipeline_id=14912d93-4ab1-4baa-826e-61752c52ed51) started.

PR #105, opened 2026-04-28
Fix: Package misses required source files

ReinierMaas:
XRef: https://github.com/channable/opsqueue/pull/50/changes#r2745964330

Adding the `includes` broke using the package downstream, preventing us from updating our OpsQueue pin.

- [x] Verified this works downstream (Nix querying is remarkably slow so pushing this to the background for now)

---

Cachix had an outage:

[image](https://status.cachix.org/786464712)


ReinierMaas:
@Qqwy is taking over this PR because he wants to simplify the packaging story further.

PR #109, opened 2026-05-23
Fix clippy recipe to fail on warnings

### LINKED ISSUE #106: Cargo Clippy succeeds with warnings if --fix is passed
In `Run just-lint` I spotted warnings that weren't denied even though we are running `caro clippy` with `-Dwarnings` / `--deny warnings`:

`cargo clippy --no-deps --fix --allow-dirty --allow-staged -- -Dwarnings`

The issue is that `clippy` decides to exit with status code `0` on warnings if `--fix` is supplied.
We should consider also running the lint without the `--fix` to actually fail on warnings.

nanookclaw:
Closes #106.

This keeps the existing `cargo clippy --fix` pass, then runs clippy again without `--fix`. That second pass preserves the `-Dwarnings` exit behavior, so `lint-heavy` fails when warnings remain instead of accepting them after cargo fix.

Running the stricter pass exposed existing Python binding error-shape lints. I marked the few rich error-returning methods with the same local allow style already used for type complexity, adding `result_large_err` where needed. Cargo also refreshed the two local package versions in `Cargo.lock` from 0.35.1 to 0.35.2 to match `Cargo.toml`.

Verification:
- `just --show clippy`
- `just --dry-run clippy`
- `cargo fmt --check`
- `cargo clippy --no-deps -- -Dwarnings`
- `just clippy`

I also ran `just lint-heavy`; clippy passed, then mypy stopped on missing local Python dependencies/stubs (`cbor2`, `multiprocess`, `opentelemetry` exporter) before code-specific type errors.


ReinierMaas:
Thanks for your contribution!

We retain a strictly linear history for this repository so I am going to rebase your PR on master instead of retaining the Merge latest master commit.

PR #111, opened 2026-05-28
Bump pyo3: 0.25.1 -> 0.28.3

### LINKED ISSUE #36: Upgrade PyO3 (0.23.x -> 0.27.x)
Opsqueue's Python library uses PyO3 to generate FFI bindings.

The PyO3 project is evolving fast, with a lot of fixes and improvements happening in the past year.
However, there are some changes to the API / Rust macros, which means that some of the PyO3 annotations will need to be rewritten to be  compatible with the newer version.

**An important reason to perform this upgrade soon, is so we can support Python 3.14.x**

Steps:

- [x] Blindly try to upgrade and see how much things break
- [x] Open up one or more follow-up issues depending on how much work this turns out to be.

jerbaroo:
This PR includes 3 commits. For each commit simply bumped the minor version of pyo3 then addressed the compiler deprecation warnings and errors and ensured that the test-suite still passed.

@Qqwy while a ✅ from the compiler and test-suite are good signs, I'm not entirely confident whether the code in `async_util` is what we want, would suggest review keenly 🙏🏻

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2618,9 +2618,9 @@ dependencies = [
 
 [[package]]
 name = "pyo3"
-version = "0.25.1"
+version = "0.26.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8970a78afe0628a3e3430376fc5fd76b6b45c4d43360ffd6cdd40bdde72b682a"
+checksum = "7ba0117f4212101ee6544044dae45abe1083d30ce7b29c4b5cbdfa2354e07383"
 dependencies = [
  "chrono",
  "indoc",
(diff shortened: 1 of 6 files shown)


ReinierMaas (approved):
To me this makes sense but I haven't written much `pyo3` code.

jerbaroo:
@OpsBotPrime merge

OpsBotPrime:
Rebased as 002dfd6eed922a753544bb3c098956a687de026f, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/43fe0acc-3187-4532-be15-cc1aed74e6cd?pipeline_id=605e43bf-f15f-4229-aa2d-2b397ed4bf90) started.

PR #112, opened 2026-06-08
Lookup submision IDs by strategic metadata

### LINKED ISSUE #64: Support checking whether a submission currently exists in the queue for a particular piece of strategic metadata
In certain use-cases it can be useful to check whether a submission for that particular situation already exists. (And if it does, Producer code might for example decide to keep waiting for that submission to finish, or instead to cancel that one and start a new one).

Instead of requiring people to store that information separately from Opsqueue, we already have the strategic metadata available.

So the goal of this task is to introduce a new function to the ProducerClient, which you can give one (or multiple, if it isn't too much effort) key-values of strategic metadata, which returns a list of in-progress submissions matching that set of metadata.

For example, you could ask whether there are any submissions for `user_id: 42`.

jerbaroo:
This PR adds the `lookup_submission_ids_by_strategic_metadata` method to `ProducerClient`, which will return the submission IDs of in-progress subsmissions that match ALL of the provided strategic metadata key-value pairs.

```py
    def lookup_submission_ids_by_strategic_metadata(
        self, strategic_metadata: dict[str, int]
    ) -> list[SubmissionId]:
        """Attempts to find in-progress submissions where the strategic metadata
        of that submission includes all of the key-value pairs of the given
        'strategic_metadata'. A matching submission must include all of the
        given key-value pairs, but it may also contain other key-value pairs.

        Raises:
        - `TooManyMatchingSubmissionsError` if the lookup matches more
          submissions than the server's configured maximum. Narrow the query
          with more specific strategic metadata.
        - `InternalProducerClientError` if there is a low-level internal error.

        """
```

### DIFF
--- a/libs/opsqueue_python/python/opsqueue/producer.py
+++ b/libs/opsqueue_python/python/opsqueue/producer.py
@@ -18,6 +18,7 @@ from opsqueue.exceptions import (
     SubmissionFailedError,
     SubmissionNotCancellableError,
     SubmissionNotFoundError,
+    TooManyMatchingSubmissionsError,
 )
 from .opsqueue_internal import (  # type: ignore[import-not-found]
     SubmissionId,
@@ -38,6 +39,7 @@ __all__ = [
     "SubmissionNotCancellable",
     "SubmissionNotCancellableError",
     "SubmissionNotFoundError",
+    "TooManyMatchingSubmissionsError",
     "ChunkFailed",
 ]
 
@@ -366,6 +368,25 @@ class ProducerClient:
         - `InternalProducerClientError` if there is a low-level internal error.
         """
         return self.inner.lookup_submission_id_by_prefix(prefix)
+
+    def lookup_submission_ids_by_strategic_metadata(
+        self, strategic_metadata: dict[str, int]
+    ) -> list[SubmissionId]:
+        """Attempts to find in-progress submissions where the strategic metadata
+        of that submission includes all of the key-value pairs of the given
+        'strategic_metadata'. A matching submission must include all of the
+        given key-value pairs, but it may also contain other key-value pairs.
+
+        Raises:
+        - `TooManyMatchingSubmissionsError` if the lookup matches more
+          submissions than the server's configured maximum. Narrow the query
+          with more specific strategic metadata.
+        - `InternalProducerClientError` if there is a low-level internal error.
+
+        """
+        return self.inner.lookup_submission_ids_by_strategic_metadata(  # type: ignore[no-any-return]
+            strategic_metadata
+        )
 
     def is_completed(self, submission_id: SubmissionId) -> bool:
         raise NotImplementedError
--- a/libs/opsqueue_python/src/errors.rs
+++ b/libs/opsqueue_python/src/errors.rs
@@ -5,7 +5,7 @@ use std::error::Error;
 use opsqueue::common::chunk::ChunkId;
 use opsqueue::common::errors::{
     ChunkNotFound, E, IncorrectUsage, SubmissionNotCancellable, SubmissionNotFound,
-    UnexpectedOpsqueueConsumerServerResponse,
+    TooManyMatchingSubmissions, UnexpectedOpsqueueConsumerServerResponse,
 };
 use pyo3::exceptions::PyBaseException;
 use pyo3::{Bound, PyErr, Python, import_exception};
@@ -22,6 +22,7 @@ import_exception!(opsqueue.exceptions, TryFromIntError);
 import_exception!(opsqueue.exceptions, ChunkNotFoundError);
 import_exception!(opsqueue.exceptions, SubmissionNotFoundError);
 import_exception!(opsqueue.exceptions, SubmissionNotCancellableError);
+import_exception!(opsqueue.exceptions, TooManyMatchingSubmissionsError);
 import_exception!(opsqueue.exceptions, NewObjectStoreClientError);
 import_exception!(opsqueue.exceptions, SubmissionNotCompletedYetError);
 
@@ -143,6 +144,12 @@ impl From<CError<SubmissionNotFound>> for PyErr {
     fn from(value: CError<SubmissionNotFound>) -> Self {
         let submission_id = value.0.0;
         SubmissionNotFoundError::new_err(u64::from(submission_id))
+    }
+}
+
+impl From<CError<TooManyMatchingSubmissions>> for PyErr {
+    fn from(value: CError<TooManyMatchingSubmissions>) -> Self {
+        TooManyMatchingSubmissionsError::new_err(value.0.0)
     }
 }
 
--- a/libs/opsqueue_python/tests/test_roundtrip.py
+++ b/libs/opsqueue_python/tests/test_roundtrip.py
@@ -14,6 +14,7 @@ from opsqueue.producer import (
     SubmissionNotFoundError,
     SubmissionNotCancellable,
     SubmissionNotCancellableError,
+    TooManyMatchingSubmissionsError,
 )
 from opsqueue.consumer import ConsumerClient, Chunk
 from opsqueue.common import SerializationFormat
@@ -21,11 +22,11 @@ from conftest import (
     background_process,
     multiple_background_processes,
     OpsqueueProcess,
+    opsqueue_service,
     StrategyDescription,
     strategy_from_description,
 )
 import logging
-
 import pytest
 
 
@@ -577,3 +578,92 @@ def test_failed_submission_includes_chunks_done(opsqueue: OpsqueueProcess) -> No
         with pytest.raises(SubmissionFailedError) as exc_info:
             producer_client.blocking_stream_completed_submission(submission_id)
         assert exc_info.value.submission.chunks_done == len(chunks) - 1
+
+
+def test_lookup_submission_ids_by_strategic_metadata(opsqueue: OpsqueueProcess) -> None:
+    """Lookup of submission IDs should only match in progress submissions with
+    all pieces of strategic metadata.
+
+    """
+    url = "file:///tmp/opsqueue/test_lookup_submission_ids_by_strategic_metadata"
+    producer_client = ProducerClient(f"localhost:{opsqueue.port}", url)
+    id_1 = producer_client.insert_submission(
+        [1], chunk_size=1, strategic_metadata={"foo": 1, "bar": 2, "wow": 3}
+    )
+    id_2 = producer_client.insert_submission(
+        [1], chunk_size=1, strategic_metadata={"foo": 1, "bar": 2, "moo": 3}
+    )
+    # Inserting some similar data to that above, which shouldn't get matched.
+    producer_client.insert_submission(
+        [1], chunk_size=1, strategic_metadata={"foo": 2, "bar": 1}
+    )
+
+    def test_lookup(
+        strategic_metadata: dict[str, int], expected_ids: list[int]
+    ) -> None:
+        found_ids = producer_client.lookup_submission_ids_by_strategic_metadata(
+            strategic_metadata
+        )
+        assert isinstance(found_ids, list)
+        assert all(map(lambda x: isinstance(x, SubmissionId), found_ids))
+        assert found_ids == expected_ids
+
+    test_lookup({"foo": 1}, [id_1, id_2])
+    test_lookup({"foo": 1, "bar": 2}, [id_1, id_2])
+    test_lookup({"foo": 1, "MISS": 2}, [])
+    test_lookup({"wow": 3}, [id_1])
+
+    # Should only match in-progress submission.
+    producer_client.cancel_submission(id_1)
+    test_lookup({"foo": 1}, [id_2])
+
+
+def test_lookup_submission_ids_by_empty_strategic_metadata(
+    opsqueue: OpsqueueProcess,
+) -> None:
+    """Lookup of submission IDs with empty strategic_metadata should NOT raise
+    an exception.
+
+    """
+    url = "file:///tmp/opsqueue/test_lookup_submission_ids_by_empty_strategic_metadata"
+    producer_client = ProducerClient(f"localhost:{opsqueue.port}", url)
+    count = 6
+    for _ in range(count):
+        producer_client.insert_submission([1], chunk_size=1)
+    assert len(producer_client.lookup_submission_ids_by_strategic_metadata({})) == count
+
+
+def test_lookup_too_many_submission_ids_by_strategic_metadata() -> None:
+    """Lookup of too many submission IDs beyond the configured limit raises
+    TooManyMatchingSubmissionsError.
+
+    """
+    max_ = 2
+    # We didn't request the OpsQueueProcess as a parameter so an instance isn't
+    # started, instead we start one here with custom args.
+    with opsqueue_service(
+        command_args=["--max-submissions-returned", str(max_)]
+    ) as opsqueue:
+        url = "file:///tmp/opsqueue/test_lookup_too_many_matching_submissions"
+        producer_client = ProducerClient(f"localhost:{opsqueue.port}", url)
+        inserted: list[SubmissionId] = []
+        strategic_metadata = {"k": 1}
+        for _ in range(max_ + 1):
+            assert (
+                inserted
+                == producer_client.lookup_submission_ids_by_strategic_metadata(
+                    strategic_metadata
+                )
+            )
+            inserted.append(
+                producer_client.insert_submission(
+                    [1], chunk_size=1, strategic_metadata=strategic_metadata
+                )
+            )
+        with pytest.raises(TooManyMatchingSubmissionsError) as exc:
+            assert len(inserted) == max_ + 1
+            producer_client.lookup_submission_ids_by_strategic_metadata(
+                strategic_metadata
+            )
+        assert exc.type is TooManyMatchingSubmissionsError
+        assert exc.value.max_submissions == max_
--- a/opsqueue/src/common/mod.rs
+++ b/opsqueue/src/common/mod.rs
@@ -1,5 +1,6 @@
 //! Common datatypes and errors shared across all parts of Opsqueue
 use rustc_hash::FxHashMap;
+use std::num::NonZero;
 
 pub mod chunk;
 pub mod errors;
@@ -12,3 +13,50 @@ pub mod submission;
 /// consider hashing them and using that hash as MetaStateVal.
 pub type MetaStateVal = i64;
 pub type StrategicMetadataMap = FxHashMap<String, MetaStateVal>;
+
+/// Maximum number of submissions a lookup may return.
+/// Guarantees: 0 < MaxSubmissions 1 < i64::MAX;
+#[derive(Debug, Clone, Copy)]
+pub struct MaxSubmissions(NonZero<u64>);
+
+impl MaxSubmissions {
+    pub fn new(value: NonZero<u64>) -> Result<Self, MaxSubmissionsTooLarge> {
+        if u64::from(value) < i64::MAX as u64 {
+            Ok(Self(value))
+        } else {
+            Err(MaxSubmissionsTooLarge(value))
+        }
+    }
+}
+
+impl From<MaxSubmissions> for u64 {
+    fn from(max_submissions: MaxSubmissions) -> u64 {
+        u64::from(max_submissions.0)
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("max_submissions value {0} is too large; it must be at most i64::MAX - 1")]
+pub struct MaxSubmissionsTooLarge(pub NonZero<u64>);
+
+impl std::fmt::Display for MaxSubmissions {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ParseMaxSubmissionsError {
+    #[error(transparent)]
+    NotANumber(#[from] std::num::ParseIntError),
+    #[error(transparent)]
+    TooLarge(#[from] MaxSubmissionsTooLarge),
+}
+
+impl std::str::FromStr for MaxSubmissions {
+    type Err = ParseMaxSubmissionsError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let value: NonZero<u64> = s.parse()?;
+        Ok(MaxSubmissions::new(value)?)
+    }
+}
--- a/opsqueue/src/common/submission.rs
+++ b/opsqueue/src/common/submission.rs
@@ -270,15 +270,17 @@ impl Submission {
 pub mod db {
     use crate::{
         common::{
-            StrategicMetadataMap,
-            errors::{DatabaseError, E, SubmissionNotCancellable, SubmissionNotFound},
+            MaxSubmissions, StrategicMetadataMap,
+            errors::{
+                DatabaseError, E, SubmissionNotCancellable, SubmissionNotFound,
+                TooManyMatchingSubmissions,
+            },
         },
         db::{Connection, True, WriterConnection, WriterPool},
     };
-    use chunk::ChunkSize;
-    use sqlx::{Sqlite, query};
-
     use axum_prometheus::metrics::{counter, histogram};
+    use chunk::ChunkSize;
+    use sqlx::{QueryBuilder, Sqlite, query, query_scalar};
 
     use super::*;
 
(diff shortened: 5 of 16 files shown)


ReinierMaas:
Open question shouldn't an empty `StrategicMetadataMap` return all submissions instead of raising an error on that?
Also this is an easy way to DDoS OpsQueue if the number of returned submissions is large.

libs/opsqueue_python/python/opsqueue/producer.py
```diff
@@ -367,6 +371,25 @@ def lookup_submission_id_by_prefix(self, prefix: str) -> SubmissionId | None:
...
+        self, strategic_metadata: dict[str, int]
+    ) -> list[SubmissionId]:
+        """Attempts to find in-progress submissions where the strategic metadata
+        of that submission includes all of the key-value pairs of the given
+        'strategic_metadata'. A matching submission must include all of the
+        given key-value pairs, but it may also contain other key-value pairs.
+
+        Raises:
+        - `LookupIdsWithEmptyStrategicMetadataError` if the provided
+          'strategic_metadata' contained no key-value pairs to look for.
```
ReinierMaas: Can this potentially also raise:

```
- `InternalProducerClientError` if there is a low-level internal error.
```

libs/opsqueue_python/python/opsqueue/producer.py
```diff
@@ -367,6 +371,25 @@ def lookup_submission_id_by_prefix(self, prefix: str) -> SubmissionId | None:
         """
         return self.inner.lookup_submission_id_by_prefix(prefix)
 
+    def lookup_submission_ids_by_strategic_metadata(
+        self, strategic_metadata: dict[str, int]
+    ) -> list[SubmissionId]:
```
ReinierMaas, line 374: I checked the other APIs on the producer and they either provide a single element or an iterator so that it can be lazily evaluated and doesn't need to be materialized in memory all at once.

opsqueue/src/producer/server.rs
```diff
@@ -133,6 +139,16 @@ async fn lookup_submission_id_by_prefix(
     Ok(Json(submission_id))
 }
 
+async fn lookup_submission_ids_by_strategic_metadata(
+    State(state): State<ServerState>,
+    extract::Json(strategic_metadata): extract::Json<StrategicMetadataMap>,
+) -> Result<Json<Vec<SubmissionId>>, ServerError> {
```
ReinierMaas: Our `SubmissionId`s are close to `u64::max` in serialized size, which is reasonable for our random (snowflake) identifiers. This issues the following serialized to JSON response size:

* 1 (`[`) + elements * 20 + elements (`,|]`) characters

For ~1 million this is ~20MB. We can put that as an configurable upper bound to return to the client with an explicit error if it goes over that so that people can update the settings and return the larger response knowingly.

libs/opsqueue_python/python/opsqueue/producer.py
ReinierMaas: I would change the interface to raise an explicit error if there are too many matching Submissions, we need that for DoS protection anyway, but allow an empty `strategic_metadata` set to be processed.

jerbaroo:
> Open question shouldn't an empty StrategicMetadataMap return all submissions instead of raising an error on that?

It depends on the API contract we want to provide. An argument in favour of throwing an error is that if a caller is using an API to "search for submissions matching some strategic metadata" but they haven't provided any strategic metadata, then they possibly have an issue in their code that they should be alerted to. Also, the SQLite query will need a special case to handle this, more code for something we don't have evidence for that it is a usecase we want to support.

ReinierMaas (commented):
I have a few comment and would like to point out that for the other queries we also assert the query plan. This allows us to spot potentially very bad execution behaviour.

libs/opsqueue_python/python/opsqueue/producer.py
jerbaroo: Yes, good spot. The rust producer/client.rs code lifts network and decoding errors into this type. Documented in b557df0b1c2f2f57bf2db2c60da3be98d2456b95

jerbaroo: SGTM. Added support for empty `strategic_metadata` in 8057405. Will add the streaming / DoS protection in separate commits.

libs/opsqueue_python/python/opsqueue/producer.py
jerbaroo: The other `ProducerClient` APIs which are returning iterators are returning _chunk_ iterators. These chunks are deterministically-addressable. Internally, the iterator is keeping track of a submission prefix, a current index, and a max index, then on each call to `next()` the iterator is performing a network request (`GET <object-storage-path><prefix>/<current-index>-out.bin`) to fetch the chunk.

This new API is different (not chunks, not querying object storage) from the existing APIs that return iterators, so would need some new wiring put in place to support streaming. But since we are only dealing with submission IDs, it's less of a memory concern. A pragmatic solution for now might be to just implement the configurable upper bound + add some metrics and keep an eye on it.

opsqueue/src/producer/server.rs
jerbaroo: Configurable upper bound added in 54bc50bb974199124ffd4f53e136b1e4d4ad81ee with a roundtrip test in be5f6d2db02efddabab29228f63ef75d64c4be5a

jerbaroo: It might be a bit over-engineered with a newtype that ensures we don't overflow when adding (+1) when building the SQLite query, but.. types good right ? 😅

jerbaroo:
> I have a few comment and would like to point out that for the other queries we also assert the query plan. This allows us to spot potentially very bad execution behaviour.

Query plan asserted in c6c9f04fa8c96f4041d6e82838b39c3b42d96dbf

justfile
ReinierMaas, line 1: I am not sure these timeouts are long enough on CI.

I had the same experience locally and made it configurable for local use:

https://github.com/channable/opsqueue/pull/122/changes#diff-deb9bb56fb122db0b605aa5b63f95a4665c905b18dd670e1fa6c877576a94ff1

libs/opsqueue_python/src/errors.rs
```diff
@@ -146,6 +147,12 @@ impl From<CError<SubmissionNotFound>> for PyErr {
     }
 }
 
+impl From<CError<TooManyMatchingSubmissions>> for PyErr {
+    fn from(value: CError<TooManyMatchingSubmissions>) -> Self {
+        TooManyMatchingSubmissionsError::new_err(value.0 .0)
```
ReinierMaas:
```suggestion
        TooManyMatchingSubmissionsError::new_err(value.0.0)
```

libs/opsqueue_python/python/opsqueue/producer.py
```diff
@@ -38,10 +39,15 @@
     "SubmissionNotCancellable",
     "SubmissionNotCancellableError",
     "SubmissionNotFoundError",
+    "TooManyMatchingSubmissionsError",
     "ChunkFailed",
 ]
 
 
+class LookupIdsWithEmptyStrategicMetadataError(Exception):
+    pass
```
ReinierMaas: Is this error still produced?

opsqueue/src/common/mod.rs
```diff
@@ -12,3 +12,50 @@ pub mod submission;
 /// consider hashing them and using that hash as MetaStateVal.
 pub type MetaStateVal = i64;
 pub type StrategicMetadataMap = FxHashMap<String, MetaStateVal>;
+
+/// Maximum number of submissions a lookup may return.
+/// Constructor checks: MaxSubmissions + 1 <= i64::MAX;
```
ReinierMaas: We should also prevent 0 from appearing, i.e. use `NonZeroU64` to store the value.

libs/opsqueue_python/tests/test_roundtrip.py
```diff
@@ -508,3 +509,80 @@ def consume(x: int) -> int | None:
...
+        producer_client = ProducerClient(f"localhost:{opsqueue.port}", url)
+        inserted = 0
+        for _ in range(max_ + 1):
+            producer_client.insert_submission(
+                [1], chunk_size=1, strategic_metadata={"k": 1}
+            )
+            inserted += 1
+        with pytest.raises(TooManyMatchingSubmissionsError):
+            assert inserted == max_ + 1  # Make sure Exception wasn't raised too early.
+            producer_client.lookup_submission_ids_by_strategic_metadata({"k": 1})
```
ReinierMaas: We can also test the endpoint keeps functioning until the one that brings it over the edge, this prevents off-by-one errors:

```suggestion
        inserted = []
        for _ in range(max_ + 1):
            assert producer_client.lookup_submission_ids_by_strategic_metadata({"k": 1}) == inserted
            inserted.append(producer_client.insert_submission(
                [1], chunk_size=1, strategic_metadata={"k": 1}
            ))
        with pytest.raises(TooManyMatchingSubmissionsError):
            assert len(inserted) == max_ + 1  # Make sure Exception wasn't raised too early.
            producer_client.lookup_submission_ids_by_strategic_metadata({"k": 1})
```

opsqueue/src/common/submission.rs
```diff
@@ -1075,6 +1146,38 @@ pub mod test {
...
+          submissions_metadata as s0
+          INNER JOIN submissions on submissions.id = s0.submission_id
+        WHERE
+          s0.metadata_key = ?
+          AND s0.metadata_value = ?
+        ORDER BY
+          s0.submission_id
+        LIMIT
+          ?
+        ");
```
ReinierMaas, line 1154: This idea turned out to be so beautiful! It makes it easy to review the generated queries.

opsqueue/src/common/submission.rs
```diff
@@ -550,6 +552,74 @@ pub mod db {
...
+    ) -> QueryBuilder<'_, Sqlite> {
+        let mut query_builder: QueryBuilder<Sqlite> =
+            QueryBuilder::new("SELECT s0.submission_id FROM submissions_metadata as s0");
+        // Inner join for each piece of strategic metadata.
+        for i in 1..strategic_metadata.len() {
+            query_builder.push(format!(
+                " JOIN submissions_metadata AS s{i} ON s{i}.submission_id = s0.submission_id"
+            ));
+        }
+        query_builder.push(" INNER JOIN submissions on submissions.id = s0.submission_id");
```
ReinierMaas: This is to scope it to the `InProgress` submissions. We could make this `INNER JOIN` table configurable based on the requested `SubmissionState` enum.

opsqueue/src/common/submission.rs
```diff
@@ -550,6 +552,74 @@ pub mod db {
...
+        }
+        let mut query_builder: QueryBuilder<Sqlite> =
+            lookup_ids_by_strategic_metadata_query(&strategic_metadata, limit);
+        let rows = query_builder.build().fetch_all(conn.get_inner()).await?;
+        error_if_too_many(
+            max_submissions,
+            rows.into_iter()
+                .map(|row| row.get("submission_id"))
+                .collect(),
+        )
```
ReinierMaas: We can simplify this a bit further to make it easier to follow:

```suggestion
        let ids = if strategic_metadata.is_empty() {
            // The main query to match on strategic_metadata will fail at run-time
            // if strategic_metadata is empty, so we handle the empty case here.
            query_scalar!(
                r#"SELECT id AS "id: SubmissionId" FROM submissions ORDER BY id LIMIT ?"#,
                limit
            )
            .fetch_all(conn.get_inner())
            .await?
        } else {
            let mut query_builder: QueryBuilder<Sqlite> =
                lookup_ids_by_strategic_metadata_query(&strategic_metadata, limit);
            query_builder.build_query_scalar().fetch_all(conn.get_inner()).await?
        };
        if ids.len() as u64 > u64::from(max_submissions) {
            // If 'max_submissions' results is exceeded return an explicit error.
            Err(E::R(TooManyMatchingSubmissions(u64::from(max_submissions))))
        } else {
            Ok(ids)
        }
```

---

Note that if we add a `SubmissionState` selector we need to switch out the `FROM submissions` with the correct `submission(_*)?` tables.

opsqueue/src/common/submission.rs
```diff
@@ -550,6 +552,74 @@ pub mod db {
...
+    }
+
+    /// The query in 'lookup_ids_by_strategic_metadata', extracted for testing.
+    pub fn lookup_ids_by_strategic_metadata_query(
+        strategic_metadata: &StrategicMetadataMap,
+        limit: i64,
+    ) -> QueryBuilder<'_, Sqlite> {
+        let mut query_builder: QueryBuilder<Sqlite> =
+            QueryBuilder::new("SELECT s0.submission_id FROM submissions_metadata as s0");
```
ReinierMaas: In the other query you added `AS "id: SubmissionId"` don't we need that here?

ReinierMaas (commented):
I have a few code simplifications and one API generalisation.

opsqueue/src/common/submission.rs
jerbaroo: The first query is verified at compile-time by sqlx, it looks at the DB schema and sees an `i64`, so we need to let sqlx know that it should read this column as `SubmissionId`. 

The second query is not verified at compile-time, we are just constructing a string and saying "trust me, it will return SubmissionIds at run-time"

justfile
jerbaroo: Great! Reverted the timeout in this PR in  9a85c035304a3d9f28ce6dfe9bc72ca366d2b066

libs/opsqueue_python/src/errors.rs
jerbaroo: Fixed in e09a1ea1e334997295ce7f116b7744e10966bc64

libs/opsqueue_python/python/opsqueue/producer.py
jerbaroo: No, removed in 6d27f09a6ca98f6cbfdc09224f9570d17d435e96

opsqueue/src/common/mod.rs
jerbaroo: Switched to `NonZero<u64>` in 441d2ca6e722d6ef9a5f4e06992dc3f92405d393

libs/opsqueue_python/tests/test_roundtrip.py
jerbaroo: Good idea, accepted in e84bbc449c17b0b30a2c642ee1d65f688018173f

opsqueue/src/common/submission.rs
jerbaroo: I like the idea, but would add it in a separate PR.

opsqueue/src/common/submission.rs
jerbaroo: Nice simplification! Adopted in cbb1dd0948df0cc0771d57975268ab0170190777

opsqueue/src/common/submission.rs
jerbaroo: Tracking in issue https://github.com/channable/opsqueue/issues/127

libs/opsqueue_python/tests/test_roundtrip.py
```diff
@@ -508,3 +509,87 @@ def consume(x: int) -> int | None:
...
+            inserted.append(
+                producer_client.insert_submission(
+                    [1], chunk_size=1, strategic_metadata={"k": 1}
+                )
+            )
+        with pytest.raises(TooManyMatchingSubmissionsError):
+            assert (
+                len(inserted) == max_ + 1
+            )  # Make sure Exception wasn't raised too early.
+            producer_client.lookup_submission_ids_by_strategic_metadata({"k": 1})
```
ReinierMaas: I remember something about `pytest.raises` not always catching the error you expect and needing to have additional checks:

https://github.com/channable/imaginator/blob/75a752a630bacca36127824019d86e3aee152535/integration/frida/test_frida_extraction.py#L53-L59

opsqueue/src/common/submission.rs
```diff
@@ -544,6 +546,63 @@ pub mod db {
...
+        max_submissions: MaxSubmissions,
+        mut conn: impl Connection,
+    ) -> Result<Vec<SubmissionId>, E<DatabaseError, TooManyMatchingSubmissions>> {
+        // MaxSubmissions provides us with the guarantee this won't overflow.
+        let limit = (u64::from(max_submissions) + 1) as i64;
+        // The main query to match on strategic_metadata will fail at run-time
+        // if strategic_metadata is empty, so we handle the empty case here.
+        let ids = if strategic_metadata.is_empty() {
+            query_scalar!(
+                r#"SELECT id AS "id: SubmissionId" FROM submissions ORDER by id LIMIT ?"#,
```
ReinierMaas:
```suggestion
                r#"SELECT id AS "id: SubmissionId" FROM submissions ORDER BY id LIMIT ?"#,
```

opsqueue/src/common/submission.rs
```diff
@@ -1069,6 +1129,38 @@ pub mod test {
         );
     }
 
+    #[sqlx::test]
+    pub async fn test_query_plan_lookup_by_strategic_metadata(db: sqlx::SqlitePool) {
+        let mut conn = db.acquire().await.unwrap();
+        let strategic_metadata: StrategicMetadataMap =
+            [("company_id".to_string(), 1), ("company_id".to_string(), 2)]
```
ReinierMaas: If you change one of the two stategic metadata keys to a different string this would cause us to see the recursion you wrote above.

opsqueue/src/common/submission.rs
```diff
@@ -544,6 +546,63 @@ pub mod db {
...
+        }
+    }
+
+    /// The query in 'lookup_ids_by_strategic_metadata', extracted for testing.
+    pub fn lookup_ids_by_strategic_metadata_query(
+        strategic_metadata: &StrategicMetadataMap,
+        limit: i64,
+    ) -> QueryBuilder<'_, Sqlite> {
+        let mut query_builder: QueryBuilder<Sqlite> =
+            QueryBuilder::new("SELECT s0.submission_id FROM submissions_metadata as s0");
```
ReinierMaas: Can we instead start of with the `submissions` table?

opsqueue/src/common/submission.rs
```diff
@@ -544,6 +546,63 @@ pub mod db {
...
+        // Inner join for each piece of strategic metadata.
+        for i in 1..strategic_metadata.len() {
+            query_builder.push(format!(
+                " JOIN submissions_metadata AS s{i} ON s{i}.submission_id = s0.submission_id"
```
ReinierMaas: Then we can start from the `0` index here instead of skipping that.

opsqueue/src/common/submission.rs
```diff
@@ -544,6 +546,63 @@ pub mod db {
...
+            ));
+        }
+        query_builder.push(" INNER JOIN submissions on submissions.id = s0.submission_id");
```
ReinierMaas: And we can drop this line from the query altogether.

ReinierMaas (commented):
I have some questions about the query we build-up. I don't think I fully follow without the recursive example working. And I expect that I hit on a small simplification.

libs/opsqueue_python/tests/test_roundtrip.py
jerbaroo: added an additional assert on the type of exception and also checking it's inner value. All changes for this review round in 76d4ede

opsqueue/src/common/submission.rs
jerbaroo: Fixed in 76d4ede and also applied the query simplification from your other comments where we select directly from `submissions` and avoid the additional join.

opsqueue/src/common/submission.rs
```diff
@@ -544,6 +546,62 @@ pub mod db {
...
+            ));
+        }
+        for (i, (key, value)) in strategic_metadata.iter().enumerate() {
+            query_builder.push(if i == 0 { " WHERE" } else { " AND" });
+            query_builder.push(format!(" s{i}.metadata_key = "));
+            query_builder.push_bind(key.clone());
+            query_builder.push(format!(" AND s{i}.metadata_value = "));
+            query_builder.push_bind(*value);
+        }
+        query_builder.push(" ORDER BY s0.submission_id LIMIT ");
```
ReinierMaas, line 598: We can simplify this a bit further to a single for loop:

```suggestion
        // Inner join for each piece of strategic metadata.
        for (i, (key, value)) in strategic_metadata.iter().enumerate() {
            query_builder.push(format!(
                " INNER JOIN submissions_metadata AS s{i} ON s{i}.submission_id = submissions.id"
            ));
            query_builder.push(format!(" AND s{i}.metadata_key = "));
            query_builder.push_bind(key.clone());
            query_builder.push(format!(" AND s{i}.metadata_value = "));
            query_builder.push_bind(*value);
        }
        query_builder.push(" ORDER BY s0.submission_id LIMIT ");
```

opsqueue/src/common/submission.rs
```diff
@@ -1069,6 +1128,42 @@ pub mod test {
...
+          JOIN submissions_metadata AS s1 ON s1.submission_id = submissions.id
+        WHERE
+          s0.metadata_key = ?
+          AND s0.metadata_value = ?
+          AND s1.metadata_key = ?
+          AND s1.metadata_value = ?
+        ORDER BY
+          s0.submission_id
+        LIMIT
+          ?
```
ReinierMaas, line 1153: Which changes the query to the following grouped statements:

```suggestion
        SELECT
          id
        FROM
          submissions
          INNER JOIN submissions_metadata AS s0 ON s0.submission_id = submissions.id
          AND s0.metadata_key = ?
          AND s0.metadata_value = ?
          INNER JOIN submissions_metadata AS s1 ON s1.submission_id = submissions.id
          AND s1.metadata_key = ?
          AND s1.metadata_value = ?
        ORDER BY
          s0.submission_id
        LIMIT
          ?
```

ReinierMaas (commented):
LGTM! I was still thinking about a last small improvement to simplify the query generation code but I am not married to that.

ReinierMaas (approved):

ReinierMaas:
The linter still failed on CI:

```diff
$ git diff
diff --git a/libs/opsqueue_python/tests/test_roundtrip.py b/libs/opsqueue_python/tests/test_roundtrip.py
index 2251c32..1f6ac50 100644
--- a/libs/opsqueue_python/tests/test_roundtrip.py
+++ b/libs/opsqueue_python/tests/test_roundtrip.py
@@ -582,13 +582,19 @@ def test_lookup_too_many_submission_ids_by_strategic_metadata() -> None:
         for _ in range(max_ + 1):
             assert (
                 inserted
-                == producer_client.lookup_submission_ids_by_strategic_metadata(strategic_metadata)
+                == producer_client.lookup_submission_ids_by_strategic_metadata(
+                    strategic_metadata
+                )
             )
             inserted.append(
-                producer_client.insert_submission([1], chunk_size=1, strategic_metadata=strategic_metadata)
+                producer_client.insert_submission(
+                    [1], chunk_size=1, strategic_metadata=strategic_metadata
+                )
             )
         with pytest.raises(TooManyMatchingSubmissionsError) as exc:
             assert len(inserted) == max_ + 1
-            producer_client.lookup_submission_ids_by_strategic_metadata(strategic_metadata)
+            producer_client.lookup_submission_ids_by_strategic_metadata(
+                strategic_metadata
+            )
         assert exc.type is TooManyMatchingSubmissionsError
         assert exc.value.max_submissions == max_

```

opsqueue/src/common/submission.rs
jerbaroo: Yay! Included in 23ed08c1b042804c394ddd0f5ae82f0f5494687f

jerbaroo:
@OpsBotPrime merge and tag

OpsBotPrime:
Your merge request has been denied, because merging on Fridays is not recommended. To override this behaviour use the command `merge and tag on Friday`.

jerbaroo:
@OpsBotPrime merge and tag on friday

OpsBotPrime:
Rebased as dcca94da91f71207e47eea2a045a58d3b5084336, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/8bad3d4e-302e-4329-b9f7-c3345097bd22?pipeline_id=5c9744fb-19fc-4a40-9d11-9e596d7faf18) started.

OpsBotPrime:
The [build failed :x:](https://channable.semaphoreci.com/workflows/8bad3d4e-302e-4329-b9f7-c3345097bd22?pipeline_id=5c9744fb-19fc-4a40-9d11-9e596d7faf18).

If this is the result of a flaky test, then tag me again with the `retry` command.  Otherwise, push a new commit and tag me again.

jerbaroo:
@OpsBotPrime retry

OpsBotPrime:
Your merge request has been denied, because merging on Fridays is not recommended. To override this behaviour use the command `retry on Friday`.

jerbaroo:
@OpsBotPrime retry on friday

OpsBotPrime:
Rebased as 931173e388b776fc3faaeb5f5184fe950b400826, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/4d33bf51-0d59-4343-9636-22d6c4494b97?pipeline_id=17bc8cc3-c00a-4773-84ed-5a1a5716cd69) started.

OpsBotPrime:
@jerbaroo I tagged your PR with [v57](https://github.com/channable/opsqueue/releases/tag/v57). Please wait for the build of 931173e388b776fc3faaeb5f5184fe950b400826 to pass and don't forget to deploy it!

PR #113, opened 2026-06-17
[CRE-1299] - support white listed requests

gallone2000:


opsqueue/src/common/submission.rs
```diff
@@ -268,12 +272,18 @@ pub mod db {
...
 
     use axum_prometheus::metrics::{counter, histogram};
 
     use super::*;
 
+    fn normalize_strategic_metadata(
+        strategic_metadata: Option<sqlx::types::Json<StrategicMetadataMap>>,
+    ) -> Option<StrategicMetadataMap> {
+        strategic_metadata.and_then(|json| (!json.0.is_empty()).then_some(json.0))
+    }
```
ReinierMaas: As requested offline:

I think it is better to keep the empty JSON logic. We shouldn't return a different value than the submitter entered.

I.e. with the current implementation one could choose to always insert submissions with a body, even if it is the empty JSON object. On requesting the chunks it can blankly parse the returned value as JSON because this is the application contract. This function would change that existing behaviour.

opsqueue/src/common/submission.rs
ReinierMaas, line 1: I would like to point out that for the other queries we also assert the query plan. This allows us to spot potentially very bad execution behaviour. Could you add that for these queries?

opsqueue/src/common/submission.rs
```diff
@@ -1046,6 +1075,11 @@ pub mod test {
             .expect("insertion failed");
 
         let fetched_submission = get_submission(submission.id, &mut conn).await.unwrap();
+        // When fetched from DB with no metadata rows, json_group_object returns '{}'.
```
ReinierMaas, line 1222: Ah so there is no distinction between `Option<StrategicMetadataMap>` and `StrategicMetadataMap` on the reader side?

---

Reading a bit further indeed shows to me that we can also remove the `Option` from the type to make this clear, this is also the type contract on `InsertSubmission`. Only the outer function `insert_submission_chunks` accepts the `Option<StrategicMetadataMap>` but it is `unwrap_or_default` into an empty map there.

ReinierMaas (commented):
LGTM! I have two optional extensions, one is removing the type level distinction that doesn't actually exist. I think that would make the whole API easier to use and understand!

opsqueue/src/common/submission.rs
gallone2000: it makes sense

opsqueue/src/common/.submission.rs.pending-snap
ReinierMaas, line 1: What is this file? Is it a alternative storage format for the `assert_snapshot!` tests? We don't have the same for the tests in `strategy.rs`.

opsqueue/src/common/submission.rs
```diff
@@ -493,7 +493,10 @@ pub mod db {
                 chunks_done: row.chunks_done,
                 chunk_size: row.chunk_size,
                 metadata: row.metadata,
-                strategic_metadata: row.strategic_metadata.map(|json| json.0),
+                strategic_metadata: row
+                    .strategic_metadata
+                    .map(|json| json.0)
+                    .unwrap_or_default(),
```
ReinierMaas: I think `unwrap_or_default` can be prevented by adding an exclamation mark at 479:

`AS "strategic_metadata!: sqlx::types::Json<StrategicMetadataMap>"`

> All of these functions return a valid JSON or JSONB object or array, even if there are no valid input rows. ~ https://sqlite.org/json1.html#jgroupobject

There shouldn't be any cases where `json_group_object` doesn't return a valid JSON object.

opsqueue/src/common/submission.rs
```diff
@@ -424,7 +424,7 @@ pub mod db {
             chunks_done: ChunkCount::zero(),
             chunk_size,
             metadata,
-            strategic_metadata: None,
+            strategic_metadata: strategic_metadata.clone(),
```
ReinierMaas: This clone can be prevented as `insert_submission` below now can rely on the `strategic_metadata` travelling with the submission. Internally it only uses references to the data so duplicating it can be prevented completely.

ReinierMaas (changes requested):
I require a few more changes.

opsqueue/src/common/.submission.rs.pending-snap
gallone2000: This is an auto-generated temporary file from insta for inline snapshot updates in submission.rs (not an alternative snapshot storage).
I accidentally added it; I’ll remove it and keep the accepted snapshot changes directly in submission.rs.

opsqueue/src/common/submission.rs
```diff
@@ -1041,11 +1214,16 @@ pub mod test {
...
+        insert_submission(submission.clone(), chunks, &mut conn)
             .await
             .expect("insertion failed");
 
         let fetched_submission = get_submission(submission.id, &mut conn).await.unwrap();
+        // When fetched from DB with no metadata rows, json_group_object returns '{}'.
+        let submission = Submission {
+            strategic_metadata: Default::default(),
+            ..submission
+        };
```
ReinierMaas, line 1226: The submission now always carries the `strategic_metadata` and it is by default the default so this should be unnecessary.

opsqueue/src/common/submission.rs
```diff
@@ -461,25 +468,37 @@ pub mod db {
...
+            Some(row) => Ok(Submission {
+                id: row.id,
+                prefix: row.prefix,
+                chunks_total: row.chunks_total,
+                chunks_done: row.chunks_done,
+                chunk_size: row.chunk_size,
+                metadata: row.metadata,
+                strategic_metadata: row.strategic_metadata.0,
+                otel_trace_carrier: row.otel_trace_carrier,
+            }),
```
ReinierMaas, line 501: Can we now with the `AS "strategic_metadata!: sqlx::types::Json<StrategicMetadataMap>"` switch back to the shorter `query_as!` without the manual mapping? Or is the `row.strategic_metadata.0` the only syntax that prevents us from doing so?

ReinierMaas (approved):
LGTM! I have two small remaining notes but nothing blocking the merge.

gallone2000:
@OpsBotPrime merge and deploy to production

OpsBotPrime:
Unknown or invalid command found:

    comment:1:30:
      |
    1 | @OpsBotPrime merge and deploy to production
      |                              ^
    No deployment environments have been configured.
[Basic usage is explained here.](https://github.com/channable/hoff/blob/master/readme.md#using-hoff)

gallone2000:
@OpsBotPrime merge and deploy

OpsBotPrime:
Unknown or invalid command found:

    comment:1:30:
      |
    1 | @OpsBotPrime merge and deploy
      |                              ^
    No deployment environments have been configured.
[Basic usage is explained here.](https://github.com/channable/hoff/blob/master/readme.md#using-hoff)

gallone2000:
@OpsBotPrime merge and tag

OpsBotPrime:
Rebased as 7572ac64df14b60a8078213a038fb87d8ce3ee3c, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/9a803077-3456-4254-b77e-7cd7e3a6ba1a?pipeline_id=db7012f6-d07e-4307-b071-e6d995df9412) started.

OpsBotPrime:
@gallone2000 I tagged your PR with [v53](https://github.com/channable/opsqueue/releases/tag/v53). Please wait for the build of 7572ac64df14b60a8078213a038fb87d8ce3ee3c to pass and don't forget to deploy it!

Issue #48, opened 2026-01-29
Feature: Client receives asynchronous `ChunkReservationExpired` not actionable from library


ReinierMaas:
https://github.com/channable/opsqueue/blob/bd0a42e7f45ccd2750071e3dd3cd9bd4dda4fd25/opsqueue/src/consumer/client.rs#L293-L295

I would expect a local cancellation token that we can await to be informed we should stop processing the `Chunk`.

PR #122, opened 2026-07-09
Increment retries on tried Chunks

ReinierMaas:
XRef: https://github.com/channable/opsqueue/issues/48

This roundtrips via the client to prevent network issues from incrementing the retry count.

### DIFF
--- a/justfile
+++ b/justfile
@@ -2,6 +2,10 @@
 _default:
   just --list --unsorted
 
+# Timeout (seconds) for Python integration test recipes.
+# Override with OPSQUEUE_PYTEST_TIMEOUT, e.g. OPSQUEUE_PYTEST_TIMEOUT=600 just nix-test-integration
+pytest_timeout_seconds := env_var_or_default("OPSQUEUE_PYTEST_TIMEOUT", "60")
+
 # Build-and-run the opsqueue binary (development profile)
 [group('run')]
 run *OPSQUEUE_ARGS:
@@ -44,10 +48,12 @@ test-unit *TEST_ARGS:
 test-integration *TEST_ARGS: build-bin build-python
   #!/usr/bin/env bash
   set -euo pipefail
+  export OPSQUEUE_BIN="$PWD/target/debug/opsqueue"
+
   cd libs/opsqueue_python
   source "./.setup_local_venv.sh"
 
-  timeout 600 pytest --color=yes {{TEST_ARGS}}
+  timeout {{pytest_timeout_seconds}} pytest --color=yes {{TEST_ARGS}}
 
 # Python integration test suite, using artefacts built through Nix. Args are forwarded to pytest
 [group('nix')]
@@ -55,13 +61,14 @@ nix-test-integration *TEST_ARGS: nix-build-bin
   #!/usr/bin/env bash
   set -euo pipefail
   nix_build_python_library_dir=$(just nix-build-python)
+  nix_build_bin_dir=$(just nix-build-bin)
 
   cd libs/opsqueue_python/tests
   export PYTHONPATH="${nix_build_python_library_dir}/lib/python3.13/site-packages"
-  export OPSQUEUE_VIA_NIX=true
+  export OPSQUEUE_BIN="${nix_build_bin_dir}/bin/opsqueue"
   export RUST_LOG="opsqueue=debug"
 
-  timeout 600 pytest --color=yes {{TEST_ARGS}}
+  timeout {{pytest_timeout_seconds}} pytest --color=yes {{TEST_ARGS}}
 
 # Run all linters, fast and slow
 [group('lint')]
--- a/justfile
+++ b/justfile
@@ -81,6 +81,10 @@ lint-heavy: clippy mypy
 clippy:
   cargo clippy --no-deps --fix --allow-dirty --allow-staged -- -Dwarnings
   cargo clippy --no-deps -- -Dwarnings
+  cargo clippy --no-deps --fix --allow-dirty --allow-staged --no-default-features -- -Dwarnings
+  cargo clippy --no-deps --no-default-features -- -Dwarnings
+  cargo clippy --no-deps --fix --allow-dirty --allow-staged --all-features -- -Dwarnings
+  cargo clippy --no-deps --all-features -- -Dwarnings
 
 # Python static analysis type-checker
 [group('lint')]
--- a/libs/opsqueue_python/pyproject.toml
+++ b/libs/opsqueue_python/pyproject.toml
@@ -54,4 +54,4 @@ addopts = "-n 4 --dist=worksteal"
 # Individual tests should be very fast. They should never take multiple seconds
 # If after 120sec (accomodating for a toaster-like PC, or an overloaded Semaphore runner)
 # there is no progress, assume a deadlock
-# timeout=120
+timeout=20
--- a/libs/opsqueue_python/tests/conftest.py
+++ b/libs/opsqueue_python/tests/conftest.py
@@ -67,6 +67,8 @@ def opsqueue_service(
     # temp_dbname = f"/tmp/opsqueue_tests-{uuid.uuid4()}.db"
 
     command = [
+        "setpriv",
+        "--pdeathsig=SIGKILL",
         str(opsqueue_bin_location()),
         "--port",
         str(port),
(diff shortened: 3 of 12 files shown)


Qqwy:
The main reason why a reservation expiration exists, is to deal with consumers that are hanging. If they are stuck in an infinite loop with the OS not being overloaded, everything is fine as the Rust async executor in the client can then still make forward progress. But if the consumer machine really has a hard time, we cannot be 100% sure they will respond to a "yo your reservation has expired" message.

Note that issues that clearly are network related (graceful and ungraceful 'network closed' as well as 'too many heartbeat failures') _do_ result in the retry count being incremented. I _think_ this is the right choice generally speaking, since it will prevent a submission from ever being infinitely stuck inside opsqueue all-together.

So you might be able to simplify this PR further, as I don't think we need to do the round-trip.

(I'm perfectly fine with merging this first and simplifying it after if that has your preference though).

ReinierMaas:
I think it makes sense to have different paths for a client that is not responding to health-checks, i.e. it has actual network issues. On the other in the case we can respond to network requests we ensure that the reservation expiration is pushed behind the chunk completion if that already entered the network.

This also was easy to generalize to the reserved chunks failure path for which the requester disappeared so that is a nice addition I would say.

---

I would also be fine incrementing the retry count completely server side if that has your preference but the client is responding to health-checks in this case so we can assume it is making forward progress. We have a separate issue to add a cancellation token to the Chunk so that the requester can now that the chunk it is currently processing has expired, we just haven't implemented that and I think it would make sense to do that in the future as well. XRef: https://github.com/channable/opsqueue/issues/48

libs/opsqueue_python/tests/conftest.py
```diff
@@ -31,11 +31,8 @@ class OpsqueueProcess:
 
 @functools.cache
 def opsqueue_bin_location() -> Path:
-    if os.environ.get("OPSQUEUE_VIA_NIX"):
-        deriv_path = (
-            subprocess.check_output(["just", "nix-build-bin"]).decode("utf-8").strip()
-        )
-        return Path(deriv_path) / "bin" / "opsqueue"
+    if explicit_bin := os.environ.get("OPSQUEUE_BIN", "").strip():
+        return Path(explicit_bin)
```
ReinierMaas, line 35: This was required to reduce the time spent in the setup and teardown fixtures so we can enable test timeouts. The binaries are now computed once in the Just file for all the `xdist` workers.

justfile
```diff
@@ -2,6 +2,10 @@
 _default:
   just --list --unsorted
 
+# Timeout (seconds) for Python integration test recipes.
+# Override with OPSQUEUE_PYTEST_TIMEOUT, e.g. OPSQUEUE_PYTEST_TIMEOUT=600 just nix-test-integration
+pytest_timeout_seconds := env_var_or_default("OPSQUEUE_PYTEST_TIMEOUT", "60")
```
ReinierMaas, line 7: Local test timeout is now set at a more reasonable 1 minute instead of the 10 minutes required on CI...
Also the build phase for the binaries is excluded from this timeout to prevent network retries in Nix realise and Cargo fetch from triggering the test timeout.

justfile
```diff
@@ -80,6 +87,8 @@ lint-heavy: clippy mypy
 [group('lint')]
 clippy:
   cargo clippy --no-deps --fix --allow-dirty --allow-staged -- -Dwarnings
+  cargo clippy --no-deps --fix --allow-dirty --allow-staged --no-default-features -- -Dwarnings
+  cargo clippy --no-deps --fix --allow-dirty --allow-staged --all-features -- -Dwarnings
```
ReinierMaas, line 93: I was reading through the open PRs and found https://github.com/channable/opsqueue/pull/109
This showed that we need to extend this even further to make cargo clippy actually fail on warnings, now it only tries to resolve them using `--fix` and won't exit with a status code for remaining warnings.

libs/opsqueue_python/tests/test_roundtrip.py
```diff
@@ -400,7 +469,7 @@ def test_cancel_in_progress(
 
 
 def test_cancel_already_cancelled(
-    opsqueue: OpsqueueProcess, any_consumer_strategy: StrategyDescription
+    opsqueue: OpsqueueProcess,
 ) -> None:
```
ReinierMaas, line 473: Removing the unused strategies reduced the number of tests that are executed while retaining the same coverage.

libs/opsqueue_python/pyproject.toml
```diff
@@ -54,4 +54,4 @@ addopts = "-n 4 --dist=worksteal"
 # Individual tests should be very fast. They should never take multiple seconds
 # If after 120sec (accomodating for a toaster-like PC, or an overloaded Semaphore runner)
 # there is no progress, assume a deadlock
-# timeout=120
+timeout=20
```
ReinierMaas, line 57: 10s did timeout sporadically so I bumped this to 20 seconds. If it turns out not to be enough I would double it again but remain lower than the shorter timeout in the just file (60s/1 minute) so that a `timeout` timeout doesn't hinder `pytest` from printing the timeout reason locally.

opsqueue/src/consumer/dispatcher/metastate.rs
```diff
@@ -20,12 +20,17 @@ impl MetaState {
 
     pub fn decrement(&self, key: &str, val: &MetaStateVal) {
         let ripe_for_removal = {
-            let meta_state_field = self
-                .0
-                .get(key)
-                .expect("decrements should be paired with increments.");
-            meta_state_field.decrement(val);
-            meta_state_field.is_empty()
+            if let Some(meta_state_field) = self.0.get(key) {
```
jerbaroo, line 23: Might be worth adding a comment why this might occur in practice, something like:

```rs
The decrement is called unconditionally after `finish_reservation`, regardless of whether `finish_reservation` succeeded or not. In other words, the `decrement` is not part of the same atomic state transition as removing the reservation from the reserver. This means that if a reservation is finished by multiple different racing paths we can get multiple decrements.
```

opsqueue/src/consumer/client.rs
```diff
@@ -158,6 +156,50 @@ fn retry_policy() -> impl BackoffBuilder {
         .without_max_times()
 }
 
+#[allow(clippy::type_complexity)]
+#[derive(Debug, Clone)]
+struct InFlightRequests(
```
jerbaroo, line 171: Can you add a short docstring explaining what this is, something like: "A shared tuple of: next request nonce and a mapping of nonce to channel (where we expect replies to that request)".

opsqueue/src/consumer/client.rs
```diff
@@ -282,16 +325,42 @@ impl Client {
...
+                                                }
+                                            },
+                                        }
                                     },
                                     ServerToClientMessage::Async(msg) => {
                                         match msg {
                                             AsyncServerToClientMessage::ChunkReservationExpired(chunk_id) => {
                                                 tracing::info!("Server indicated that we took too long with {chunk_id:?}, and now our reservation has expired.");
+                                                // Send message to the server to indicate that we are no longer reserving this chunk, so that it can be re-reserved by other consumers.
+                                                // This round-trip is necessary to avoid a race condition where we sent the success or failure of the chunk to the server, while we lost the reservation window.
```
jerbaroo: I think it would be worth adding to the comment what the effect of that race condition would be, something like: "In a race where a chunk success was already sent to the server, this change allows that chunk to succeed and not be spuriously retried (connection ordering guarantees the round-trip of the reservation expiry will come after the success)."

jerbaroo (approved):
LGTM!

It would be worth describing the updated expiration/retry pathway in the top-level README.

Pushing the reservation expiry to the consumer, who then sends a fail chunk, sounds good to me because it allows for richer logic to be added consumer-side in response to the expiry (e.g. #48)

ReinierMaas:
@OpsBotPrime merge and tag!

OpsBotPrime:
Rebased as 0ede7e1c9862b6a1cfbd984df5c9fbff5bc77a7c, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/e66bf541-255c-43cd-b083-1affcedd36e2?pipeline_id=59f4e601-0f20-4389-830f-f40eccaacdea) started.

OpsBotPrime:
Abandoning this pull request because it was closed.

ReinierMaas:
@OpsBotPrime merge and tag!

OpsBotPrime:
Rebased as 4b8c7d7daa9655af6d5dc1e75dded4a1cbffe66e, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/ff39a7f2-e422-471e-8383-9944836757d7?pipeline_id=f3cecd5b-0ec6-4448-a9bd-007c74456eba) started.

OpsBotPrime:
@ReinierMaas I tagged your PR with [v56](https://github.com/channable/opsqueue/releases/tag/v56). Please wait for the build of 4b8c7d7daa9655af6d5dc1e75dded4a1cbffe66e to pass and don't forget to deploy it!

PR #128, opened 2026-07-14
Add Jobmachine delegation

SemMulder:


SemMulder:
Discussed offline with @ReinierMaas: currently fetching chunks to work on is a bottleneck in their system. Adding paused the way we do here would make the issue worse.

Therefore:
 - Instead of tracking the submission paused state with a column in `submissions`, follow the partitioning of `submissions` on status (i.e. create a `submissions_paused`).
 - Similarly, partition the `chunks` table on status (i.e. `chunks_paused`, `chunks_completed`, `chunks_failed`, etc.).

PR #132, opened 2026-07-15
Implement pausing for submissions

SemMulder:
Introduce `submissions_paused` and `chunks_paused` tables
(alongside the existing `submissions_{completed,failed,cancelled}` and
`chunks_{completed,failed}` tables).

A submission can now be created in a Paused state. It's then stored in
`submissions_paused` and its chunks are stored in `chunks_paused`.
Because paused chunks are not in the `chunks` table, the consumer
dispatcher naturally skips them without any changes to the dispatch
query.

Unpausing moves the submission and the chunks to `submissions` and
`chunks` and notifies waiting consumers.

Paused submissions are cancellable; `cancel_submission` now handles
the case where the submission is found in `submissions_paused`.

We don't allow pausing submissions after creation. That proved to
have too many edge cases we would need to resolve.

Also cleans up some code encountered along the way.


SemMulder:
TODO: How to deal with reserved chunks when pausing?

SemMulder:
> TODO: How to deal with reserved chunks when pausing?

Discussed this with @ReinierMaas yesterday, we tried to make it work s.t. chunk completions for cancelled and paused submissions would still be processed. However, implementing that proved to be difficult:
 - What if the completed chunk was the last one? Would the submission status then progress from `cancelled`/`paused` to `completed`?
 - There were a lot of subtleties that needed to be thought through around multiple consumers that could be working on the same chunk.

After struggling with it for a while, I thought it would be easier to invoke the idempotency assumption we have for how consumers process chunks, and just ignore the completion in case the submission is cancelled or paused. This results in less edge-cases and simpler code in the end, at the cost of slightly higher compute cost. But I think it's worth the trade-off: there are enough edge-cases in the codebase as it is :).

@ReinierMaas, let me know if you disagree ;).

libs/opsqueue_python/python/opsqueue/exceptions.py
```diff
@@ -92,15 +92,6 @@ class TryFromIntError(IncorrectUsageError):
     pass
 
 
-class ChunkNotFoundError(IncorrectUsageError):
```
SemMulder, line 95: Note that `ChunkNotFoundError` can be dropped since it is (and was) dead code.

opsqueue/src/common/chunk.rs
```diff
@@ -281,25 +281,18 @@ pub mod db {
...
+        conn.transaction(move |mut tx| {
+            Box::pin(async move {
+                complete_chunk_raw(chunk_id, output_content, &mut tx).await?;
+                crate::common::submission::db::maybe_complete_submission(
+                    chunk_id.submission_id,
+                    &mut tx,
+                )
+                .await
             })
-            .await;
```
SemMulder, line 322: Note that we were silently dropping errors here before, which went unnoticed because of the underscore prefix of `_chunk_size`.

opsqueue/src/common/submission.rs
```diff
@@ -406,16 +427,174 @@ pub mod db {
...
+                    )
+                    .await?;
+                    super::chunk::db::insert_many_paused_chunks(&chunks, &mut tx).await?;
+                    Ok(())
+                }
+                .boxed()
+            })
+            .await;
+
+        counter!(crate::prometheus::SUBMISSIONS_TOTAL_COUNTER).increment(1);
```
SemMulder, line 483: TODO: Increment `SUBMISSIONS_PAUSED_COUNTER`.

opsqueue/src/common/submission.rs
```diff
@@ -1644,4 +1943,240 @@ pub mod test {
...
+            true,
+            &mut conn,
+        )
+        .await
+        .expect("insertion failed");
+
+        assert_matches!(count_submissions(&mut conn).await, Ok(0));
+        assert_matches!(count_submissions_paused(&mut conn).await, Ok(1));
+        assert_matches!(count_submissions_completed(&mut conn).await, Ok(0));
+    }
```
SemMulder: TODO: Drop this test, it doesn't add much compared to the tests above.

SemMulder:
> After struggling with it for a while, I thought it would be easier to invoke the idempotency assumption we have for how consumers process chunks, and just ignore the completion in case the submission is cancelled or paused. 

Thought some more: if the assumption is that a paused job should not have any work being done, this will violate that. Maybe easier to just only allow pausing of not yet started submissions for now?

Other option is to make pausing async, and only transition from `InProgress` to `Paused` after the pending chunks have been completed/failed. But that will likely blow up this PR. TBD.

SemMulder:
> > After struggling with it for a while, I thought it would be easier to invoke the idempotency assumption we have for how consumers process chunks, and just ignore the completion in case the submission is cancelled or paused.
> 
> Thought some more: if the assumption is that a paused job should not have any work being done, this will violate that. Maybe easier to just only allow pausing of not yet started submissions for now?
> 
> Other option is to make pausing async, and only transition from `InProgress` to `Paused` after the pending chunks have been completed/failed. But that will likely blow up this PR. TBD.

After sleeping on this for a night, I'm leaning towards ignoring the problem, getting this PR merged, and opening a new ticket to resolve this in the future. Otherwise this PR will become too big IMO (and it is already quite hefty).

Will discuss with @ReinierMaas.

ReinierMaas:
I thought a bit on it. I think the easiest way is to only allow a DAG:

```
active (paused -> running) -> completed (cancelled | failed | success)
```

We can insert a submissions in either `paused` or `running` and then only move forward from there to keep it as simple as possible. 

From `paused` we can move to any of the `completed` states, say we don't have any chunks that is an automatic completed successfully. I am not sure how we can move from `paused` to `failed`, maybe you have an example for that? We can move from `paused` to `cancelled` by cancelling the submission instead of `unpausing`. If a `paused` submissions is `unpaused` it moves to `running`.

From `running` we can move to to any of the `completed` states:
* no chunks or all chunks succeeded -> `success`
* cancellation of the submission -> `cancelled`
* a chunk exceeds its retries -> `failed`

If we need more complicated logic we can pick that up as a separate issue.

---

To be abundantly clear how the DAG state transitions would look like:

```mermaid
graph LR
    %% Active States
    P((Paused))
    R((Running))

    %% Completed States
    S([Success])
    F([Failed])
    C([Cancelled])

    %% Internal Active Transitions
    P -- unpause --> R

    %% Transitions to Completed States
    P -- no chunks --> S
    P -.-> F
    P -- user cancellation --> C

    R -- no chunks / all succeed --> S
    R -- chunk retries exceeded --> F
    R -- user cancellation --> C
```

SemMulder:
> I am not sure how we can move from paused to failed

I think this is impossible, it would always go via `running`.

> We can insert a submissions in either paused or running and then only move forward from there to keep it as simple as possible.

This would indeed make it simpler. In practice, that would amount to removing the `pause`  endpoint from this PR. For our current use-cases this would be fine. That is, we currently only need the ability to submit `paused` submissions, we don't need to `pause` them mid-way.

Note that this does mean that we will have to implement `/job/return` as a no-op once we start implementing [JM Delegation](https://github.com/channable/jobmachine/blob/ae7bab76002084def8a592581e2c3771cf25b786/docs/jobmachine-delegation/worker_requirements.md#jobreturn). However, I believe that's fine. From there:

> This request is non-binding and can be ignored for tasks that have started running already.

But will double-check that with @radekchannable as well.

SemMulder:
> But will double-check that with @radekchannable as well.

Double-checked, we indeed don't *have* to pause anything when JM asks.

SemMulder:
> > I am not sure how we can move from paused to failed

> I think this is impossible, it would always go via `running`.

This was wrong :sweat_smile:, you can cancel a paused submission, then it would end up in failed.

libs/opsqueue_python/src/producer.rs
```diff
@@ -376,57 +377,6 @@ impl ProducerClient {
...
-            .map_err(|CError(e)| {
-                CError(match e {
-                    L(e) => L(e),
-                    R(L(e)) => R(L(e)),
-                    R(R(e)) => R(R(R(e))),
-                })
-            })?;
-        Ok(res)
-    }
-
```
SemMulder, line 429: This was dead code, because it is re-implemented on the Python side.

ReinierMaas: Shouldn't we then prefer the Rust side over the Python side as that is exposed to all downstream consumers.

libs/opsqueue_python/src/producer.rs
```diff
@@ -158,6 +159,36 @@ impl ProducerClient {
         })
     }
 
+    /// Unpause a paused submission, making it available to consumers again.
```
ReinierMaas:
```suggestion
    /// Unpause a paused submission, making it available to consumers.
```

opsqueue/src/prometheus.rs
```diff
@@ -66,6 +68,16 @@ pub fn describe_metrics() {
...
     );
+    describe_counter!(
+        SUBMISSIONS_PAUSED_COUNTER,
+        Unit::Count,
+        "Number of submissions paused"
+    );
+    describe_counter!(
+        SUBMISSIONS_UNPAUSED_COUNTER,
+        Unit::Count,
+        "Number of submissions unpaused (resumed)"
```
ReinierMaas: We no longer support resuming as we use a DAG, as that means running then pausing then running again.

```suggestion
        "Number of submissions unpaused"
```

libs/opsqueue_python/tests/test_roundtrip.py
```diff
@@ -667,3 +703,90 @@ def test_lookup_too_many_submission_ids_by_strategic_metadata() -> None:
...
+
+        consumer_client.run_each_op(process_op)
+
+    with background_process(run_consumer) as _consumer:
+        with pytest.raises(TimeoutError):
+            producer_client.run_submission(
+                [1],
+                chunk_size=1,
+                timeout=0.1,
+            )
```
ReinierMaas, line 727: We always check something on the exception raised by pytest, i.e. from a bit higher in the file:

```python
# We expect the intended attributes to be there:
assert isinstance(exc_info.value.failure, str)
assert isinstance(exc_info.value.submission, SubmissionFailed)
assert isinstance(exc_info.value.chunk, ChunkFailed)
```

Just checking the instance of `exc_info` would be enough here, we have experienced pytest catching the wrong exceptions and giving the green checkmark incidentally

```suggestion
        with pytest.raises(TimeoutError) as exc_info:
            producer_client.run_submission(
                [1],
                chunk_size=1,
                timeout=0.1,
            )
        assert isinstance(exc_info, TimeoutError)
```

libs/opsqueue_python/tests/test_roundtrip.py
```diff
@@ -667,3 +703,90 @@ def test_lookup_too_many_submission_ids_by_strategic_metadata() -> None:
...
+    producer_client = ProducerClient(f"localhost:{opsqueue.port}", url)
+    submission_id = producer_client.insert_submission(
+        (1, 2, 3), chunk_size=1, paused=False
+    )
+    assert isinstance(
+        producer_client.get_submission_status(submission_id),
+        SubmissionStatus.InProgress,
+    )
+    with pytest.raises(SubmissionNotFoundError):
+        producer_client.unpause_submission(submission_id)
```
ReinierMaas, line 777: Same `pytest.raises` remark here.

opsqueue/src/common/chunk.rs
```diff
@@ -583,6 +601,93 @@ pub mod db {
...
+        submission_id: SubmissionId,
+        mut conn: impl WriterConnection,
+    ) -> sqlx::Result<()> {
+        let now = chrono::prelude::Utc::now();
+
+        let query_res = sqlx::query!(
+            "
+    INSERT INTO chunks_failed
+    (submission_id, chunk_index, input_content, failure, skipped, failed_at)
+    SELECT submission_id, chunk_index, input_content, '', 1, julianday($1) FROM chunks_paused WHERE submission_id = $2;
```
ReinierMaas: https://www.sqlite.org/datatype3.html#boolean_datatype

```suggestion
    SELECT submission_id, chunk_index, input_content, '', TRUE, julianday($1) FROM chunks_paused WHERE submission_id = $2;
```

opsqueue/src/common/chunk.rs
```diff
@@ -583,6 +601,93 @@ pub mod db {
...
+                    .push_bind(chunk.chunk_index)
+                    .push_bind(chunk.input_content.clone());
+            });
+            let query = query_builder.build();
+
+            query.execute(conn.get_inner()).await?;
+        }
+
+        Ok(())
+    }
```
ReinierMaas, line 638: Funny that this is done in batches but the moves between tables are done wholesale.

I understand why we perform it a batch at a time, latency with regards to the async runtime system but I would have thought that moves between tables than also need to be performed in batches for the system as a whole to adhere to the don't block a runtime thread for a long time.

NOTE: This is not something that needs to be picked up in this PR more something that needs to be discussed in an issue.

opsqueue/src/common/submission.rs
```diff
@@ -212,12 +212,33 @@ pub struct SubmissionCancelled {
     pub cancelled_at: DateTime<Utc>,
 }
 
+/// A submission that has been paused.
+///
+/// Once a submission is paused, it gets moved to the `submissions_paused`
+/// table, and its old `submissions` record gets deleted. All remaining
+/// (non-completed, non-failed) chunks are similarly moved to `chunks_paused`.
+///
+/// A paused submission can be unpaused (resumed) or cancelled.
```
ReinierMaas: This doc coment still references the `resume` option that is not possible in the DAG.

Also if any chunk moves to failed the submission moves to failed. So it describes a flow that cannot happen.

opsqueue/src/producer/client.rs
```diff
@@ -194,6 +194,49 @@ impl Client {
         .await
     }
 
+    /// Unpause a paused submission, making it available to consumers again.
```
ReinierMaas:
```suggestion
    /// Unpause a paused submission, making it available to consumers.
```

opsqueue/src/common/submission.rs
```diff
@@ -1773,4 +2042,125 @@ pub mod test {
...
+            , chunks_done
+            , chunk_size
+            , metadata
+            , ( SELECT json_group_object(metadata_key, metadata_value)
+                FROM submissions_metadata
+                WHERE submission_id = submissions_paused.id
+              ) AS strategic_metadata
+            , otel_trace_carrier
+        FROM submissions_paused WHERE id = 1
+        ";
```
ReinierMaas: By virtue of this test not using the query that is actually used in the production code they have already diverged at the review stage...

https://github.com/channable/opsqueue/blob/9fc1444bda3f5f1a7e632199758b6b92239b6cd1/opsqueue/src/common/submission.rs#L955-L972

Can you ensure that you actually use the queries that are subsequently used in the production code-path?

opsqueue/src/common/submission.rs
```diff
@@ -429,9 +450,123 @@ pub mod db {
...
+    FROM submissions_paused WHERE id = $1;
+
+    DELETE FROM submissions_paused WHERE id = $2 RETURNING *;
+    ",
+            id,
+            id,
+        )
+        .fetch_optional(conn.get_inner())
+        .await?;
+        if row.is_none() {
```
ReinierMaas: This returns all information to just check that the `row_count == 1`. We should use `execute` and check that instead.

opsqueue/src/common/submission.rs
```diff
@@ -462,33 +598,38 @@ pub mod db {
...
+            insert_paused_submission(submission, chunks, &mut conn).await?;
+        } else {
+            insert_submission(submission, chunks, &mut conn).await?;
+            // Empty submissions get special handling: we mark them as completed right away.
+            // See https://github.com/channable/opsqueue/issues/86 for rationale.
+            if len == 0 {
+                match maybe_complete_submission(submission_id, conn).await {
+                    // Forward our database errors to the caller.
+                    Err(E::L(e)) => return Err(e),
+                    // If the submission ID can't be found, that's too bad, but it's not our problem anymore i guess.
```
ReinierMaas:
```suggestion
                    // If the submission ID can't be found, that's too bad, but it's not our problem anymore I guess.
```

opsqueue/src/common/submission.rs
```diff
@@ -579,12 +720,15 @@ pub mod db {
...
+            UNION ALL
+            SELECT id AS "id: SubmissionId" FROM submissions_completed WHERE prefix = $3
             UNION ALL
-            SELECT id AS "id: SubmissionId" FROM submissions_failed WHERE prefix = $3
+            SELECT id AS "id: SubmissionId" FROM submissions_failed WHERE prefix = $4
             "#,
             prefix,
             prefix,
+            prefix,
             prefix
```
ReinierMaas: This is a single statement query you can thus deduplicate binding the prefix. The duplicated binding only needs to happen on the multi statement queries.

opsqueue/src/common/submission.rs
```diff
@@ -900,6 +1087,22 @@ pub mod db {
...
+    ///
+    /// Returns [`SubmissionNotFound`] if the submission is not found in `submissions_paused`.
+    pub async fn cancel_paused_submission_notx(
+        id: SubmissionId,
+        mut conn: impl WriterConnection<Transaction = True>,
+    ) -> Result<(), E<DatabaseError, SubmissionNotFound>> {
+        cancel_paused_submission_raw(id, &mut conn).await?;
+        super::chunk::db::skip_remaining_paused_chunks(id, conn).await?;
+        Ok(())
+    }
```
ReinierMaas, line 1226: Should we remove `pub` it they are not intended to be called from elsewhere? Or at least scope it back to `pub(super)`/`pub(crate)`.

opsqueue/src/common/submission.rs
```diff
@@ -932,6 +1135,38 @@ pub mod db {
         }
     }
```
ReinierMaas: The same submission_opt data is unused we should use execute and rely on row_count.

opsqueue/src/common/submission.rs
```diff
@@ -932,6 +1135,38 @@ pub mod db {
...
+        if submission_opt.is_none() {
+            Err(E::R(SubmissionNotFound(id)))
+        } else {
+            counter!(crate::prometheus::SUBMISSIONS_CANCELLED_COUNTER).increment(1);
+            histogram!(crate::prometheus::SUBMISSIONS_DURATION_CANCEL_HISTOGRAM).record(
+                crate::prometheus::time_delta_as_f64(Utc::now() - id.timestamp()),
+            );
+            Ok(())
+        }
+    }
```
ReinierMaas, line 1290: The same subissions_opt data is unusued we should use execute and rely on row_count.

ReinierMaas (changes requested):
I think we are nearly there. I did find some changes we should still make but they are minor and not architectural.

libs/opsqueue_python/src/producer.rs
SemMulder: Note that this is in `opsqueue_python`, so it's not exposed to any downstream consumers ATM :).

opsqueue/src/common/submission.rs
```diff
@@ -429,9 +450,123 @@ pub mod db {
...
+    ) -> Result<(), E<DatabaseError, SubmissionNotFound>> {
+        conn.transaction(move |mut tx| {
+            Box::pin(async move {
+                unpause_submission_raw(id, &mut tx).await?;
+                super::chunk::db::restore_paused_chunks(id, &mut tx).await?;
+                Ok(())
+            })
+        })
+        .await
+    }
```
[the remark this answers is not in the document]
SemMulder: This pattern is there for all of the other metrics as well. I would suggest going for consistency in this PR, and opening an issue for solving it separately.

opsqueue/src/common/submission.rs
```diff
@@ -429,9 +450,123 @@ pub mod db {
...
+                    Ok(())
+                }
+                .boxed()
+            })
+            .await;
+
+        counter!(crate::prometheus::SUBMISSIONS_PAUSED_COUNTER).increment(1);
+        counter!(crate::prometheus::SUBMISSIONS_TOTAL_COUNTER).increment(1);
+        counter!(crate::prometheus::CHUNKS_TOTAL_COUNTER).increment(chunks_total);
+        res
```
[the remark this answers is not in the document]
SemMulder: See [the comment above](https://github.com/channable/opsqueue/pull/132#discussion_r3713450436).

libs/opsqueue_python/tests/test_roundtrip.py
SemMulder: Hmm, as far as I understand `pytest.raises(TimeoutError)` already does `isinstance(exc_info, TimeoutError)`?

As in the following should fail:

```python
with pytest.raised(NotTheRaisedError):
    raise AnotherError()
```

SemMulder: It does:

```python
>>> import pytest
>>> with pytest.raises(TimeoutError):
...     raise Exception("Not timeout error")
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
Exception: Not timeout error
>>> with pytest.raises(TimeoutError):
...     raise TimeoutError()
... 
>>> 
```

opsqueue/src/common/submission.rs
SemMulder: Pulled out all `submission_status` queries into separate functions now.

libs/opsqueue_python/tests/test_roundtrip.py
SemMulder: See my [comment above](https://github.com/channable/opsqueue/pull/132#discussion_r3713482218).

libs/opsqueue_python/src/producer.rs
```diff
@@ -158,6 +159,36 @@ impl ProducerClient {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the submission is not found or if an internal client error occurs.
+    #[allow(clippy::result_large_err, clippy::type_complexity)]
+    pub fn unpause_submission(
+        &self,
+        py: Python<'_>,
+        id: SubmissionId,
+    ) -> CPyResult<
```
ReinierMaas, line 174: Should we add the pyo3 signature like we do for `insert_submission_direct`, or is this automatically inferred?

libs/opsqueue_python/src/producer.rs
```diff
@@ -442,17 +426,39 @@ impl ProducerClient {
...
     ) -> CPyResult<
         PyChunksIter,
         E![
             FatalPythonException,
+            TryFromFloatSecsError,
+            Elapsed,
             errors::SubmissionFailed,
             InternalProducerClientError
         ],
     > {
```
ReinierMaas, line 439: Do we need to add the pyo3 signature here as well?

---

Nice addition of the timeout!

libs/opsqueue_python/tests/test_roundtrip.py
ReinierMaas: Ah nice, then we no longer need it for the outer layer and only need to use it to validate the data that is included in the exception. Thanks for confirming!

opsqueue/src/common/chunk.rs
```diff
@@ -301,27 +300,34 @@ pub mod db {
...
             })
-            .await;
+            .await?;
 
-        counter!(crate::prometheus::CHUNKS_COMPLETED_COUNTER).increment(1);
+        if chunks_moved {
+            counter!(crate::prometheus::CHUNKS_COMPLETED_COUNTER).increment(1);
+        }
         Ok(())
     }
```
ReinierMaas, line 332: Let's rename `chunks_moved` to `chunk_moved`, it became a boolean and we pass a ChunkId down. I was suprised that we update the counter with `1` and not the number of moved chunks to then find out it is a boolean.

opsqueue/src/common/chunk.rs
```diff
@@ -583,6 +607,93 @@ pub mod db {
...
+        while iter.peek().is_some() {
+            let query_chunks = iter.by_ref().take(ROWS_PER_QUERY);
+
+            let mut query_builder: QueryBuilder<Sqlite> = QueryBuilder::new(
+                "INSERT INTO chunks_paused (submission_id, chunk_index, input_content) ",
+            );
+            query_builder.push_values(query_chunks, |mut b, chunk| {
+                b.push_bind(chunk.submission_id)
+                    .push_bind(chunk.chunk_index)
+                    .push_bind(chunk.input_content.clone());
```
ReinierMaas, line 630: This is cloning the entire input contents, when in database storage is used as opposed to GCS storage backend. The pattern also exists for the other paths (i.e. master), so not something we should change in this branch, but a relative small change can circumvent this cloning:

```diff
diff --git a/opsqueue/src/common/chunk.rs b/opsqueue/src/common/chunk.rs
index c27c17e..16662c1 100644
--- a/opsqueue/src/common/chunk.rs
+++ b/opsqueue/src/common/chunk.rs
@@ -558,12 +558,12 @@ pub mod db {
     /// Returns an error if any batch insert fails.
     #[tracing::instrument(skip(chunks, conn))]
     pub async fn insert_many_chunks(
-        chunks: &[Chunk],
+        chunks: impl IntoIterator<Item = Chunk>,
         mut conn: impl WriterConnection,
     ) -> sqlx::Result<()> {
         const ROWS_PER_QUERY: usize = 1000;
 
-        let mut iter = chunks.iter().peekable();
+        let mut iter = chunks.into_iter().peekable();
         while iter.peek().is_some() {
             let query_chunks = iter.by_ref().take(ROWS_PER_QUERY);
 
@@ -573,7 +573,7 @@ pub mod db {
             query_builder.push_values(query_chunks, |mut b, chunk| {
                 b.push_bind(chunk.submission_id)
                     .push_bind(chunk.chunk_index)
-                    .push_bind(chunk.input_content.clone());
+                    .push_bind(chunk.input_content);
             });
             let query = query_builder.build();
 
diff --git a/opsqueue/src/common/submission.rs b/opsqueue/src/common/submission.rs
index 991744d..7a9fb45 100644
--- a/opsqueue/src/common/submission.rs
+++ b/opsqueue/src/common/submission.rs
@@ -418,7 +418,7 @@ pub mod db {
                         &mut tx,
                     )
                     .await?;
-                    super::chunk::db::insert_many_chunks(&chunks, &mut tx).await?;
+                    super::chunk::db::insert_many_chunks(chunks, &mut tx).await?;
                     Ok(())
                 }
                 .boxed()

```

opsqueue/src/common/chunk.rs
```diff
@@ -583,6 +607,93 @@ pub mod db {
...
+            });
+            let query = query_builder.build();
+
+            query.execute(conn.get_inner()).await?;
+        }
+
+        Ok(())
+    }
+
+    /// Move all chunks of a paused submission from `chunks_paused` back to `chunks`.
```
ReinierMaas, line 640:
```suggestion
    /// Move all chunks of a paused submission from `chunks_paused` to `chunks`.
```

opsqueue/src/common/submission.rs
```diff
@@ -430,9 +450,126 @@ pub mod db {
...
+            submission.chunk_size.0,
+        )
+        .execute(conn.get_inner())
+        .await?;
+
+        Ok(())
+    }
+
+    /// Unpause a paused submission. Atomically moves it back from `submissions_paused`
+    /// to `submissions` and its chunks from `chunks_paused` to `chunks`.
```
ReinierMaas, line 513: Maybe we should document that it atomically moves from `submissions_paused` to `submissions_completed` if it is a 0-chunk submission that is unpaused.

opsqueue/src/common/submission.rs
```diff
@@ -463,33 +601,38 @@ pub mod db {
...
+                    }
+                    // If everything went OK, this *could* still indicate a bug in producer code, so let's just log it.
+                    // Our future selves might thank us.
+                    Ok(true) => {
+                        tracing::debug!(%submission_id, "Zero-length submission marked as completed");
+                    }
+                    // This should never happen. If it does, better log it.
+                    Ok(false) => {
+                        tracing::warn!(%submission_id, "Zero-length submission wasn't zero-length?!");
+                    }
```
ReinierMaas, line 635: This logging is skipped on the unpausing path. The unpausing path does forward the `SubmissionNotFound` that is caught here? The unpausing path also does this atomically, i.e. within a transaction.

Also on all the other paths we unconditionally execute the function and here we protect it with the `if len == 0` branch... I am not completely sure how to best standardize this.

opsqueue/src/common/submission.rs
```diff
@@ -665,8 +808,124 @@ pub mod db {
     ) -> Result<Option<SubmissionStatus>, DatabaseError> {
         // NOTE: The order is important here; a concurrent writer could move a submission
         // from InProgress to Completed/Failed in-between the queries.
+        // TODO: Rewrite the queries here into a single query using `UNION ALL`.
```
ReinierMaas, line 811: Is this `TODO` something you want to still resolve? Or are you going to reference an issue `TODO(issue_nr):` before merging?

opsqueue/src/common/submission.rs
```diff
@@ -665,8 +808,124 @@ pub mod db {
     ) -> Result<Option<SubmissionStatus>, DatabaseError> {
         // NOTE: The order is important here; a concurrent writer could move a submission
         // from InProgress to Completed/Failed in-between the queries.
```
ReinierMaas, line 810:
```suggestion
// NOTE: The order is important here; a concurrent writer could move a submission
// from Paused to InProgress/Cancelled in-between the queries.
// from InProgress to Completed/Failed/Cancelled in-between the queries.
```

Also couldn't we prevent this by using an transaction and performing all the queries on the same snapshot? We are using WAL-mode:

> WAL mode permits simultaneous readers and writers. It can do this because changes do not overwrite the original database file, but rather go into the separate write-ahead log file. That means that readers can continue to read the old, original, unaltered content from the original database file at the same time that the writer is appending to the write-ahead log. In [WAL mode](https://sqlite.org/wal.html), SQLite exhibits "snapshot isolation". When a read transaction starts, that reader continues to see an unchanging "snapshot" of the database file as it existed at the moment in time when the read transaction started. Any write transactions that commit while the read transaction is active are still invisible to the read transaction, because the reader is seeing a snapshot of database file from a prior moment in time. ~ https://sqlite.org/isolation.html

opsqueue/src/common/submission.rs
```diff
@@ -1254,8 +1654,7 @@ pub mod test {
         LIMIT
           ?
         ");
-        let explained = explain_query_plan(&formatted_query, &mut conn).await;
-        assert_non_regressing_query_plan(&formatted_query, &explained);
```
ReinierMaas, line 1258: Why did you remove the `assert_non_regressing_query_plan` calls? This happened multiple times.

libs/opsqueue_python/src/producer.rs
SemMulder: In principle it is automatically inferred. The default signature here would be:

```python
def unpause_submission(id): ...
```

I can add it for consistency though.

opsqueue/src/common/submission.rs
SemMulder: I think we should move the logging of `Zero-length submission marked as completed` into `maybe_complete_submission` itself. If we then wrap the non-paused path in a transaction, we can just drop everything in the `match` here, since non of the cases will ever happen (because of snapshot isolation).

The `if len == 0` is then just an optimization in this path because we happen to have the `len` on hand. I don't mind if it stays, but we can also just drop it :shrug:.

opsqueue/src/common/submission.rs
SemMulder: Mainly as a hint for simplification in the future, will create an issue for it.

opsqueue/src/common/submission.rs
SemMulder: We could, will add that to aforementioned issue.

opsqueue/src/common/submission.rs
SemMulder: I felt it was redundant, with a snapshot check right after. But if it happened multiple times, better to have some additional guardrails.

I added them back, but as two separate functions (for clarity).

Also, the check for materialization was wrong (it checked for `MATERIALIZED`, but it should have checked for `MATERIALIZE`, see [here](https://github.com/search?q=repo%3Asqlite%2Fsqlite+%22ExplainQueryPlan%22+MATERIALIZE&type=code)).

Also used `TEMP B-TREE` as a clearer marker for temporary B-Tree construction ([proof](https://github.com/search?q=repo%3Asqlite%2Fsqlite+%22ExplainQueryPlan%22+B-TREE&type=code) that this is correct).

opsqueue/src/common/submission.rs
SemMulder: Made it nicer on the eyes now.

opsqueue/src/common/submission.rs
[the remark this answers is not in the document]
SemMulder: https://github.com/channable/opsqueue/issues/182

opsqueue/src/common/submission.rs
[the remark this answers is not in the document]
SemMulder: https://github.com/channable/opsqueue/issues/182

opsqueue/src/common/submission.rs
SemMulder: https://github.com/channable/opsqueue/issues/183

opsqueue/src/common/submission.rs
SemMulder: https://github.com/channable/opsqueue/issues/183

justfile
```diff
@@ -113,7 +113,10 @@ semver:
   set -euo pipefail
   export DATABASE_URL="sqlite://{{justfile_directory()}}/opsqueue/opsqueue_example_database_schema.db"
   # We select the latest git tag, not to be confused with the latest Cargo version.
-  cargo semver-checks --workspace --target x86_64-unknown-linux-gnu --baseline-rev "$(git tag -l --sort=-version:refname | head -1)"
+  cargo semver-checks --workspace --target x86_64-unknown-linux-gnu --baseline-rev "$(git tag -l --sort=-version:refname | head -1)" || {
+    echo "Semver checks failed. Please bump the version in Cargo.toml"
+    exit 1
+  }
```
ReinierMaas, line 119: Thanks!

opsqueue/benches/chunks_select.rs
```diff
@@ -191,6 +191,7 @@ async fn seed_or_extend(
                 None,
                 metadata,
                 ChunkSize::default(),
+                false,
```
ReinierMaas: There is a bit of boolean blindness here. Can we make it a `Paused|Started` enum?

opsqueue/src/consumer/dispatcher/mod.rs
```diff
@@ -379,6 +379,7 @@ mod test {
             None,
             StrategicMetadataMap::default(),
             ChunkSize::default(),
+            false,
```
ReinierMaas: There is a bit of boolean blindness here. Can we make it a `Paused|Started` enum?

opsqueue/src/common/chunk.rs
```diff
@@ -969,4 +1004,36 @@ pub mod test {
...
+
+        insert_submission(submission.clone(), chunks, &mut conn)
+            .await
+            .expect("insertion failed");
+
+        let res = retry_or_fail_chunk(chunk_id, "kapot".into(), &mut conn, 2).await;
+        assert_matches!(res, Ok(false));
+
+        let res = retry_or_fail_chunk(chunk_id, "kapot".into(), &mut conn, 2).await;
+        assert_matches!(res, Ok(true));
```
ReinierMaas: It was unclear to me why there was one boolean `true`, it's marked as a permanent failure here once. Can you document this behaviour?

ReinierMaas (approved):
LGTM! Thanks for the update!

Can we easily replace paused: Bool with paused: (Paused|Started/Running) enum? If not due to the Pyo3 wrapping we can leave it as is.

opsqueue/src/common/chunk.rs
SemMulder: Done!

opsqueue/src/consumer/dispatcher/mod.rs
SemMulder: Done!

opsqueue/benches/chunks_select.rs
SemMulder: Done!

SemMulder:
@OpsBotPrime merge and tag

OpsBotPrime:
Rebased as 4642e658aed4a236671fc2a0ea946048d77e7222, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/0d43cd3b-2c30-492c-87c9-35284453a800?pipeline_id=f89f6575-bf53-4c18-b031-1efa277a2920) started.

OpsBotPrime:
@SemMulder I tagged your PR with [v63](https://github.com/channable/opsqueue/releases/tag/v63). Please wait for the build of 4642e658aed4a236671fc2a0ea946048d77e7222 to pass and don't forget to deploy it!

PR #134, opened 2026-07-16
Test: Bind to port `0` for automatic port selection and report that to the fixture

### LINKED ISSUE #20: CI keeps failing due to timeout
When i open a PR, i often have to run the job a couple of times before it completes in time.

We've had issues before with our integration tests being flaky and becoming deadlocked indefinitely. #6 introduced the 20 second timeout for integration tests, so they fail more quickly once they become deadlocked.

It's also difficult to see what exactly causes the failure. A first step towards resolving this issue could be to see what we can do to improve the output that we get when a test times out on CI, because at time of writing this is just a massive stack trace with mostly callsites originating from pytest plugins and the like.

Either the time-out is just too short for CI, or we're still getting deadlocks. So far, i've not been able to really reproduce any deadlocks by running the tests locally. We could try to relax that timeout a bit more, but not by too much. As the comment above the timeout configuration states, we're already being quite generous with our time limit.

**Examples**

The last four PRs have all seen at least one failure of the integration tests due to it exceeding the time limit:

- #19 
  - [failed push](https://channable.semaphoreci.com/workflows/56fef772-2c93-4009-a632-fa1b51fe5aa4)
- #18
  - [failed push](https://channable.semaphoreci.com/jobs/ef466f0d-4f1c-4e1c-9e8f-883d47d118c5)
  - [failed re-run](https://channable.semaphoreci.com/jobs/c496f290-f42b-46ea-9605-13204f6fcae5)
- #17
  - [failed push](https://channable.semaphoreci.com/jobs/ed000520-408d-4222-a674-02833521bdf9)
- #16
  - [failed merge by Hoff](https://github.com/channable/opsqueue/pull/16#issuecomment-3376018058)

ReinierMaas:
Closes: https://github.com/channable/opsqueue/issues/20

I discovered why the tests sometimes flake and time out. The test suite by accident selected the same random port from the range and then Opsqueue couldn't bind to that port. Due to the prolonged retry behaviour for binding to the port this caused `pytest-timeout` to trigger instead of delivering us a nice error message.

### DIFF
--- a/libs/opsqueue_python/tests/conftest.py
+++ b/libs/opsqueue_python/tests/conftest.py
@@ -50,13 +50,9 @@ def opsqueue() -> Generator[OpsqueueProcess, None, None]:
 
 @contextmanager
 def opsqueue_service(
-    *, port: int | None = None
+    *,
+    port: int = 0,  # The default of 0 means "pick any free port".
 ) -> Generator[OpsqueueProcess, None, None]:
-    global test_opsqueue_port_offset
-
-    if port is None:
-        port = random_free_port()
-
     # This will create a SQLite database in memory.
     # We need the `cache=shared` to allow sharing this DB between all threads within the same OS process.
     temp_dbname = "file::memory:?cache=shared"
@@ -66,12 +62,16 @@ def opsqueue_service(
     # will from time to time hang for **many minutes** on initializing SQLite for some reason.
     # temp_dbname = f"/tmp/opsqueue_tests-{uuid.uuid4()}.db"
 
+    read_fd, write_fd = os.pipe()
+
     command = [
         "setpriv",
         "--pdeathsig=SIGKILL",
         str(opsqueue_bin_location()),
         "--port",
         str(port),
+        "--report-bound-port-pipe",
+        str(write_fd),
         "--database-filename",
         temp_dbname,
     ]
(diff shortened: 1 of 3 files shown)


libs/opsqueue_python/tests/conftest.py
```diff
@@ -50,13 +50,9 @@ def opsqueue() -> Generator[OpsqueueProcess, None, None]:
 
 @contextmanager
 def opsqueue_service(
-    *, port: int | None = None
+    *,
+    port: int = 0,
```
jerbaroo: Suggest add comment "Default 0 results in automatic port selection."

libs/opsqueue_python/tests/conftest.py
```diff
@@ -66,43 +62,47 @@ def opsqueue_service(
...
+        ) as process:
+            os.close(write_fd)
+            write_fd = -1
+
+            assert process.poll() is None, "Opsqueue process failed to start"
+            try:
+                actual_port = int.from_bytes(
+                    os.read(read_fd, 2), byteorder="big", signed=False
+                )
+                wrapper = OpsqueueProcess(port=actual_port, process=process)
```
jerbaroo: `yield OpsqueueProcess(port=actual_port, process=process)` ?

libs/opsqueue_python/tests/conftest.py
```diff
@@ -66,43 +62,47 @@ def opsqueue_service(
...
+                yield wrapper
+                assert process.poll() is None, "Opsqueue process failed during run"
+            finally:
```
jerbaroo, line 104: Suggest add a comment like: `"Exiting a subprocess.Popen context manager only results in a process.wait() which will hand indefinitely as the child server keeps running, so this is needed."`

ReinierMaas: Thanks we looked into this and it indeed only closes all the `std(in|out|err)` file descriptors but never sends a `sigterm` or `sigkill` to the managed process...

libs/opsqueue_python/tests/conftest.py
```diff
@@ -105,6 +107,18 @@ def opsqueue_service(
         os.close(read_fd)
 
 
+def read_exact_fd(fd: int, num_bytes: int) -> bytes:
```
jerbaroo, line 123: Please add a short docstring of why this is needed over just `os.read`

jerbaroo (approved):

jerbaroo:
This likely fixes https://github.com/channable/opsqueue/issues/20

ReinierMaas:
@OpsBotPrime merge!

OpsBotPrime:
Rebased as 73d79d496378d8f0785651a69e077bf21facfca2, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/f09450d0-3400-453d-b534-ce140b2d264c?pipeline_id=56629d45-dc67-4709-a2fc-df75e02678b3) started.

PR #138, opened 2026-07-16
Pass `reserver_id`s to SQLite to pre-filter the queried `chunks`

ReinierMaas:
This uses the fact that SQLite query engine lives in the same process as the calling Rust code to pass the integers directly into the query via SQL-FFI.

This can be extended to the `metastate` as well so we don't need to serialize to JSON in order to use the data in the query itself.

### DIFF
--- a/opsqueue/src/common/chunk.rs
+++ b/opsqueue/src/common/chunk.rs
@@ -124,6 +124,18 @@ impl From<ChunkIndex> for i64 {
     }
 }
 
+impl TryFrom<i64> for ChunkIndex {
+    type Error = crate::common::errors::TryFromIntError;
+
+    fn try_from(value: i64) -> Result<Self, Self::Error> {
+        if value < 0 {
+            return Err(crate::common::errors::TryFromIntError(()));
+        }
+
+        Ok(Self(u63::new(value as u64)))
+    }
+}
+
 impl TryFrom<u64> for ChunkIndex {
     type Error = crate::common::errors::TryFromIntError;
     fn try_from(value: u64) -> Result<Self, Self::Error> {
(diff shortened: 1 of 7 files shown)


opsqueue/src/common/chunk.rs
```diff
@@ -124,6 +124,18 @@ impl From<ChunkIndex> for i64 {
     }
 }
 
+impl TryFrom<i64> for ChunkIndex {
+    type Error = crate::common::errors::TryFromIntError;
+
+    fn try_from(value: i64) -> Result<Self, Self::Error> {
+        if value < 0 {
+            return Err(crate::common::errors::TryFromIntError(()));
```
jerbaroo, line 132: The error could include more information ? Such as the source value and target type?

opsqueue/src/consumer/dispatcher/mod.rs
```diff
@@ -21,6 +22,63 @@ use std::sync::Arc;
 use super::strategy;
 use crate::common::StrategicMetadataMap;
 
+unsafe extern "C" fn sqlite_reserved_chunk_lookup(
```
jerbaroo, line 25: Some documentation would be useful here, in particular about the parameters: as the parameter names and types are quite general.

opsqueue/src/consumer/strategy.rs
```diff
@@ -741,6 +799,7 @@ pub mod test {
         .unwrap();
 
         let mut conn = db_pools.reader_conn().await.unwrap();
+        register_reserved_lookup_noop(conn.get_inner()).await;
```
jerbaroo, line 802: Why is this noop being performed?

opsqueue/src/common/chunk.rs
ReinierMaas: Yes it could. We haven't done that for any of the other integer errors.

opsqueue/src/consumer/strategy.rs
ReinierMaas: If you don't register any function you can't call into SQLite to explain the plan even though you don't execute the query it checks that the called function is provided.

jerbaroo:
Added to `master` in c4376be8d369d0a25f0708b0226b2575ef24cbb4

PR #139, opened 2026-07-17
SQL: Prevent duplicate `metadata_value` for `(submission_id, metadata_key)` in `submissions_metadata`

ReinierMaas:
This brings the SQL schema inline with the code enforced `StrategicMetadataMap`.

For each submission we are tracking one `metadata_value` per key. We don't allow multiple `metadata_values` for the same key to be specified on a single submission.

PR #140, opened 2026-07-17
SQL: Order by ascending metadata counts, null first

ReinierMaas:
Extracted SQL migration to: https://github.com/channable/opsqueue/pull/139

Based on https://github.com/channable/opsqueue/pull/138 where we perform the same trick for the reserved ids.

> This uses the fact that SQLite query engine lives in the same process as the calling Rust code to pass the integers directly into the query via SQL-FFI.

This PR removes JSON serialization from the hot path to build the query to select a chunk from the datastore.


jerbaroo:
Added to `master` in 43cf3cecb0f4fade9398f742434ad712740cf2a5

PR #148, opened 2026-07-22
PreferDistinct: faster and fairer

jerbaroo:


jerbaroo:
@ReinierMaas The benchmark from this branch has been extracted into https://github.com/channable/opsqueue/pull/151, it was annoying to maintain it here, prefer to keep it separate and then it can just be cherry-picked to wherever it is needed.

jerbaroo:
Superceded by https://github.com/channable/opsqueue/pull/164

PR #150, opened 2026-07-23
CI: Double compute performance, halve build jobs

ReinierMaas:
More effective caching between the Nix builds by running in the same job.
Share feature flags between Rust builds by using `cargo hakari`.


opsqueue/src/consumer/strategy.rs
ReinierMaas, line 1: The Nix provided `sqlite3` library is of a slightly different version that the one bundled with `libsqlite-sys` crate, which causes some small differences in the explanation of the query plans.

Cargo.lock
ReinierMaas, line 1: Here you can spot the reduction in dependencies! But the workspace hack still makes sense because that consolidates the feature selection as well, which also reduces the amount of recompiles of the same (transitive) dependency.

Cargo.toml
```diff
@@ -6,11 +6,73 @@ resolver = "3"
...
+reqwest = { version = "0.13.2", default-features = false, features = ["json"] }
+rustc-hash = { version = "2.1.2" }
+sentry = { version = "0.48", default-features = false, features = ["rustls", "reqwest", "anyhow", "backtrace", "contexts", "tracing", "panic"] }
+sentry-tracing = { version = "0.48" }
+serde = { version = "1.0.203", features = ["derive"] }
+serde-error = { version = "0.1.3" }
+serde_json = { version = "1.0.149" }
+snowflaked = { version = "1.0.3", features = ["sync"] }
+sqlformat = { version = "0.5.0" }
+sqlx = { version = "0.9.0", features = ["chrono", "derive", "json", "macros", "migrate", "runtime-tokio", "sqlite", "tls-rustls-aws-lc-rs"] }
```
radekchannable: We only use SQLite, so there is no need to enable TLS support.

Cargo.toml
```diff
@@ -6,11 +6,73 @@ resolver = "3"
...
+thiserror = { version = "2.0.18" }
+tokio = { version = "1.52.1", features = ["macros", "signal", "rt-multi-thread"] }
+tokio-tungstenite = { version = "0.30.0", features = ["connect", "stream"] }
+tokio-util = { version = "0.7.18", features = ["io", "rt", "time"] }
+tower-http = { version = "0.7.0", features = ["trace", "catch-panic"] }
+tracing = { version = "0.1.41", features = ["log"] }
+tracing-opentelemetry = { version = "0.33.0" }
+tracing-subscriber = { version = "0.3", features = ["std", "env-filter", "fmt"] }
+url = { version = "2.5.2" }
```
radekchannable, line 71: By pinning `idna_adapter` to 1.1.0 this will use unicode-rs instead of ICU4X for working with Unicode domains, which should build much faster. I think we only use this for talking to GCS, Opsqueue server, and Sentry, none of which are hosted on Unicode domains, so we could even disable Unicode domain support entirely. See https://docs.rs/crate/idna_adapter/latest.

Cargo.toml
```diff
@@ -6,11 +6,73 @@ resolver = "3"
...
+opentelemetry-otlp = { version = "0.32", features = ["grpc-tonic"] }
+opentelemetry-resource-detectors = { version = "0.11.0" }
+opentelemetry-semantic-conventions = { version = "0.32.0", features = ["semconv_experimental"] }
+opentelemetry_sdk = { version = "0.32", default-features = false, features = ["trace", "rt-tokio"] }
+opsqueue = { path = "opsqueue", default-features = false }
+pyo3 = { version = "0.28.3", features = ["chrono"] }
+pyo3-async-runtimes = { version = "0.28.0", features = ["tokio-runtime", "unstable-streams"] }
+pyo3-log = { version = "0.13.3" }
+rand = { version = "0.10.1" }
+reqwest = { version = "0.13.2", default-features = false, features = ["json"] }
```
radekchannable, line 53: If we use curl instead of reqwest for talking to GCS, Opsqueue server, and Sentry that would save a lot of build time.

radekchannable (commented):
Some more suggestions but I understand if not all of them are wanted, actionable right now or at all.

Cargo.toml
ReinierMaas: > ICU4X as the default

I think this indicates that the ICU4X2.2 is indeed used.

https://github.com/channable/opsqueue/blob/fbb75f3c11dde786418ef362fe9d0dd5c4d5c308/Cargo.lock#L1402-L1410

I switched to `idna_adapter = v1.1` to use `unicode.rs`.

Cargo.toml
ReinierMaas: I am not changing the internal guts in this PR. The development experience when using `libcurl` from Rust is also worse.

radekchannable (approved):

ReinierMaas:
@OpsBotPrime merge!

OpsBotPrime:
Rebased as c239da70f4049effbcb72304fa11dbb8f51f4818, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/e452397b-c566-4c98-a0d8-28c21495c07c?pipeline_id=c2a24dd2-28ad-42a5-b24b-649d51269c32) started.

PR #151, opened 2026-07-23
Add reservation benchmark

jerbaroo:
This PR adds a benchmark which can be run with `just bench-chunks-select`. It produces the SVG contained in this PR. Output from the Rust bench is written to a CSV (not committed) which is then read by Python to produce a graph. The Rust bench also prints the information to stdout, which looks like:

```bash
SHAPE                          STRATEGY                            SIZE         MEDIAN     P10 / P90
---------------------------------------------------------------------------------------------------------
Seeding database with shape=FewSubmissionsManyChunks and total_chunks=100:
    Metadata values: 10
    Submissions per metadata value: 1
    Chunks per submission: 10
Beginning benchmark of strategy Random
FewSubmissionsManyChunks       Random                                 100          56.5       51.4 / 65.9
Beginning benchmark of strategy PreferDistinct { meta_key: "metadata_value", underlying: Oldest }
FewSubmissionsManyChunks       PreferDistinct(metadata_value, Oldest) 100          111.1      103.1 / 130.8
Seeding database with shape=FewSubmissionsManyChunks and total_chunks=500:
    Metadata values: 10
    Submissions per metadata value: 1
    Chunks per submission: 50
Beginning benchmark of strategy Random
FewSubmissionsManyChunks       Random                                 500          55.8       53.1 / 66.6
Beginning benchmark of strategy PreferDistinct { meta_key: "metadata_value", underlying: Oldest }
FewSubmissionsManyChunks       PreferDistinct(metadata_value, Oldest) 500          182.6      174.3 / 256.1
Seeding database with shape=FewSubmissionsManyChunks and total_chunks=1000:
    Metadata values: 10
    Submissions per metadata value: 1
    Chunks per submission: 100
Beginning benchmark of strategy Random
FewSubmissionsManyChunks       Random                                 1000         55.9       52.1 / 63.2
Beginning benchmark of strategy PreferDistinct { meta_key: "metadata_value", underlying: Oldest }
FewSubmissionsManyChunks       PreferDistinct(metadata_value, Oldest) 1000         344.3      260.8 / 384.3
Seeding database with shape=FewSubmissionsManyChunks and total_chunks=2000:
    Metadata values: 10
    Submissions per metadata value: 1
    Chunks per submission: 200
Beginning benchmark of strategy Random
FewSubmissionsManyChunks       Random                                 2000         55.3       50.9 / 62.7
Beginning benchmark of strategy PreferDistinct { meta_key: "metadata_value", underlying: Oldest }
FewSubmissionsManyChunks       PreferDistinct(metadata_value, Oldest) 2000         499.6      438.2 / 681.2
Seeding database with shape=FewSubmissionsManyChunks and total_chunks=4000:
    Metadata values: 10
    Submissions per metadata value: 1
    Chunks per submission: 400
Beginning benchmark of strategy Random

```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -63,6 +63,8 @@ let
         # For the shell, libpython needs to be in the search path.
         pythonEnv
       ];
+      # Fixes: libstdc++.so.6: cannot open shared object file: No such file or directory
+      LD_LIBRARY_PATH = pkgs.lib.makeLibraryPath [ pkgs.stdenv.cc.cc.lib ];
     };
   };
 in
--- a/default.nix
+++ b/default.nix
@@ -72,6 +72,8 @@ let
         # This is needed for the libsqlite3-sys crate to find the correct sqlite3.
         export PKG_CONFIG_PATH="${pkgs.sqlite.dev}/lib/pkgconfig''${PKG_CONFIG_PATH:+:''${PKG_CONFIG_PATH}}"
       '';
+      # Fixes: libstdc++.so.6: cannot open shared object file: No such file or directory
+      LD_LIBRARY_PATH = pkgs.lib.makeLibraryPath [ pkgs.stdenv.cc.cc.lib ];
     };
   };
 in
--- a/justfile
+++ b/justfile
@@ -53,7 +53,7 @@ test-integration *TEST_ARGS: build-bin build-python
   cd libs/opsqueue_python
   source "./.setup_local_venv.sh"
 
-  timeout {{pytest_timeout_seconds}} pytest --color=yes {{TEST_ARGS}}
+  pytest --timeout {{pytest_timeout_seconds}} --color=yes {{TEST_ARGS}}
 
 # Python integration test suite, using artefacts built through Nix. Args are forwarded to pytest
 [group('nix')]
@@ -68,7 +68,19 @@ nix-test-integration *TEST_ARGS: nix-build
   export OPSQUEUE_BIN="${nix_build_bin_dir}/bin/opsqueue"
   export RUST_LOG="opsqueue=debug"
 
-  timeout {{pytest_timeout_seconds}} pytest --color=yes {{TEST_ARGS}}
+  pytest --timeout {{pytest_timeout_seconds}} --color=yes {{TEST_ARGS}}
+
+# Benchmark query cost and render a graph.
+[group('bench')]
+bench-chunks-select:
+  #!/usr/bin/env bash
+  set -euo pipefail
+  cargo bench --bench chunks_select
+
+  cd libs/opsqueue_python
+  source "./.setup_local_venv.sh"
+  cd -
+  python opsqueue/benches/plot_chunks_select.py
 
 # Run all linters, fast and slow
 [group('lint')]
--- a/justfile
+++ b/justfile
@@ -75,6 +75,16 @@ nix-test-integration *TEST_ARGS: nix-build
 
   timeout --signal term --kill-after {{pytest_timeout_kill_seconds}} {{pytest_timeout_seconds}} pytest --color=yes {{TEST_ARGS}}
 
+# Benchmark query cost and render a graph.
+[group('bench')]
+bench-chunks-select:
+  #!/usr/bin/env bash
+  set -euo pipefail
+  cargo bench --bench chunks_select
+
+  source "./libs/opsqueue_python/.setup_local_venv.sh"
+  python opsqueue/benches/plot_chunks_select.py
+
 # Run all linters, fast and slow
 [group('lint')]
 lint: (lint-light "--all-files") lint-heavy
--- a/opsqueue/benches/chunks_select.rs
+++ b/opsqueue/benches/chunks_select.rs
@@ -1,0 +1,288 @@
+/// FOR EACH shape:
+///   FOR EACH amount of chunks:
+///     FOR EACH strategy:
+///
+///         // Setup
+///         Create fresh temporary SQLite database
+///         Insert chunks
+///
+///         // MEASURE
+///         FOR _ in (SAMPLES + WARMUP):
+///             Start Timer
+///             Execute SQL Query -> Fetch EXACTLY ONE chunk
+///             Stop Timer
+///             IF (not warmup): save duration
+///
+///         // REPORT & CLEANUP
+///         Calculate stats
+///         Write result to terminal and CSV
+///         Delete temporary database files
+use opsqueue::common::StrategicMetadataMap;
+use opsqueue::common::chunk::{ChunkId, ChunkSize};
+use opsqueue::common::submission::db::insert_submission_from_chunks;
+use opsqueue::consumer::dispatcher::Dispatcher;
+use opsqueue::consumer::strategy::Strategy;
+use opsqueue::db::{self};
+use std::io::Write;
+use std::num::NonZero;
+use std::path::PathBuf;
+use std::time::{Duration, Instant};
+
+// Human-readable name for an OpsQueue strategy.
+type StrategyName = &'static str;
+
+// Shape of the data we are inserting.
+#[derive(Debug, Eq, Ord, PartialEq, PartialOrd)]
+enum Shape {
+    FewSubmissionsManyChunks,
+    ManySubmissionsFewChunks,
+    Realistic,
+}
+
+const CHUNKS_PER_METADATA_VALUE: u64 = CHUNKS_PER_SUBMISSION * SUBMISSIONS_PER_METADATA_VALUE;
+const CHUNKS_PER_SUBMISSION: u64 = 1024;
+// Increase in the total amount of chunks for the `Shape::Realistic` strategy.
+const CHUNKS_STEP: usize = 500_000;
+const METADATA_VALUES: u64 = 100_000;
+const SUBMISSIONS: u64 = 300_000;
+const MAX_CHUNKS: u64 = 5_000_000;
+// The amount of samples to collect per iteration of the
+// shape/amount_of_chunks/strategy main loop. Note that each "sample" itself
+// requires a number of reservations to be performed (see `bench_strategy`).
+const SAMPLES_PER_STAT: usize = 20;
+const SUBMISSIONS_PER_METADATA_VALUE: u64 = SUBMISSIONS / METADATA_VALUES;
+// The amount of reservations to collect a single sample.
+// Note that the time between these is NOT INDEPENDENT:
+// the earlier reservations affect the later ones.
+const RESERVATIONS_IN_SAMPLE: usize = 30;
+// The amount of reservations to perform before we begin collecting a sample.
+const RESERVATIONS_IN_WARMUP: usize = 5;
+
+#[derive(Debug)]
+pub struct BenchStats {
+    pub median: f64, // Median of medians
+    pub p10: f64,    // Lower bound of medians
+    pub p90: f64,    // Upper bound of medians
+}
+
+#[allow(clippy::missing_panics_doc)]
+impl BenchStats {
+    #[must_use]
+    pub fn new(runs: Vec<Vec<f64>>) -> Self {
+        // Calculate the median for each individual run
+        let mut run_medians: Vec<f64> = runs
+            .into_iter()
+            .filter_map(|mut run_samples| {
+                if run_samples.is_empty() {
+                    None
+                } else {
+                    run_samples.sort_by(|a, b| a.partial_cmp(b).unwrap());
+                    Some(run_samples[run_samples.len() / 2])
+                }
+            })
+            .collect();
+        if run_medians.is_empty() {
+            return BenchStats {
+                p10: 0.,
+                median: 0.,
+                p90: 0.,
+            };
+        }
+        // Sort the N medians to find our distribution.
+        run_medians.sort_by(|a, b| a.partial_cmp(b).unwrap());
+        let len = run_medians.len();
+        BenchStats {
+            p10: run_medians[len * 10 / 100],
+            median: run_medians[len / 2],
+            p90: run_medians[len * 90 / 100],
+        }
+    }
+}
+
+/// Maximum amount of chunks for the bench run.
+///
+/// This is based on the shape of the data, as that can dramatically affect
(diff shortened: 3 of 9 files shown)


justfile
```diff
@@ -53,7 +53,7 @@ test-integration *TEST_ARGS: build-bin build-python
   cd libs/opsqueue_python
   source "./.setup_local_venv.sh"
 
-  timeout {{pytest_timeout_seconds}} pytest --color=yes {{TEST_ARGS}}
+  pytest --timeout {{pytest_timeout_seconds}} --color=yes {{TEST_ARGS}}
```
ReinierMaas: I don't think this covers what the previous timeout covered. The previous timeout covers the problem that the GIL isn't ever released by the OpsQueue code. This also only mentions that it is the timeout before dumping the stacks, does it also exit pytest at that point?

XRef: https://github.com/channable/opsqueue/pull/154 for dumping the transitive `xdist`-worker stacks via signal handlers on top over killing the process on hanging on the GIL lock.

ReinierMaas: We also have in python per test timeouts defined in `pyproject.toml` if that didn''t trigger it is probably a lock-up:

https://github.com/channable/opsqueue/blob/01b6b009140094dd8f8f8964e212750edea038d8/libs/opsqueue_python/pyproject.toml#L54-L57

default.nix
```diff
@@ -63,6 +63,8 @@ let
         # For the shell, libpython needs to be in the search path.
         pythonEnv
       ];
+      # Fixes: libstdc++.so.6: cannot open shared object file: No such file or directory
+      LD_LIBRARY_PATH = pkgs.lib.makeLibraryPath [ pkgs.stdenv.cc.cc.lib ];
```
ReinierMaas, line 76: Which dependency needed this? And at what time of it lifecycle build or runtime?

Just enabling the benchmarks probably caused it to show-up for you to add it here. What I am thinking is should we also ensure that our packaged `opsqueue` server has this library path available.

opsqueue/benches/plot_chunks_select.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
+"""Plot the chunks_select benchmark: one subplot per backlog shape, a line per strategy.
+
+Reads target/chunks_select_bench.csv (written by the bench) and writes a PNG next to
+this script so it can be committed alongside the code it measures.
+
+"""
```
ReinierMaas, line 7: The committed file is a SVG. Not a problem but a documentation mismatch.

opsqueue/benches/chunks_select.rs
```diff
@@ -0,0 +1,288 @@
...
+///         FOR _ in (SAMPLES + WARMUP):
+///             Start Timer
+///             Execute SQL Query -> Fetch EXACTLY ONE chunk
+///             Stop Timer
+///             IF (not warmup): save duration
+///
+///         // REPORT & CLEANUP
+///         Calculate stats
+///         Write result to terminal and CSV
+///         Delete temporary database files
```
ReinierMaas: If you flip `FOR EACH amount of chunks` and `FOR EACH strategy` you can **extend** chunks/submissions instead of generating an entirely new dataset for each measurement. This can greatly speed-up your test runs.

ReinierMaas (changes requested):
I didn't review `opsqueue/benches/chunks_select.rs` yet, only the top block comment. But I think I already have some good advice that can be helpful.

opsqueue/benches/plot_chunks_select.py
jerbaroo: Fixed the documentation to match "SVG"

default.nix
jerbaroo: This is only added to the `pkgs.mkShell`, it fixes the following error from attempting to plot the benchmark results with matplotlib:

```
Traceback (most recent call last):
  File "/home/jeremy-barisch-rooney/channable/opsqueue/opsqueue/benches/plot_chunks_select.py", line 10, in <module>
    import matplotlib.pyplot as plt  # noqa: E402
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jeremy-barisch-rooney/channable/opsqueue/libs/opsqueue_python/.venv-928fc105951d55f0ef3cc1c72c15d87e2ac058c1/lib/python3.13/site-packages/matplotlib/__init__.py", line 161, in <module>
    from . import _api, _version, cbook, rcsetup
  File "/home/jeremy-barisch-rooney/channable/opsqueue/libs/opsqueue_python/.venv-928fc105951d55f0ef3cc1c72c15d87e2ac058c1/lib/python3.13/site-packages/matplotlib/cbook.py", line 24, in <module>
    import numpy as np
  File "/home/jeremy-barisch-rooney/channable/opsqueue/libs/opsqueue_python/.venv-928fc105951d55f0ef3cc1c72c15d87e2ac058c1/lib/python3.13/site-packages/numpy/__init__.py", line 109, in <module>
    from numpy.__config__ import show_config
  File "/home/jeremy-barisch-rooney/channable/opsqueue/libs/opsqueue_python/.venv-928fc105951d55f0ef3cc1c72c15d87e2ac058c1/lib/python3.13/site-packages/numpy/__config__.py", line 4, in <module>
    from numpy._core._multiarray_umath import (
    ...<3 lines>...
    )
  File "/home/jeremy-barisch-rooney/channable/opsqueue/libs/opsqueue_python/.venv-928fc105951d55f0ef3cc1c72c15d87e2ac058c1/lib/python3.13/site-packages/numpy/_core/__init__.py", line 85, in <module>
    raise ImportError(msg) from exc
ImportError:

IMPORTANT: PLEASE READ THIS FOR ADVICE ON HOW TO SOLVE THIS ISSUE!

Importing the numpy C-extensions failed. This error can happen for
many reasons, often due to issues with your setup or how NumPy was
installed.

We have compiled some common reasons and troubleshooting tips at:

    https://numpy.org/devdocs/user/troubleshooting-importerror.html

Please note and check the following:

  * The Python version is: Python 3.13 from "/home/jeremy-barisch-rooney/channable/opsqueue/libs/opsqueue_python/.venv-928fc105951d55f0ef3cc1c72c15d87e2ac058c1/bin/python"
  * The NumPy version is: "2.5.1"

and make sure that they are the versions you expect.

Please carefully study the information and documentation linked above.
This is unlikely to be a NumPy issue but will be caused by a bad install
or environment on your machine.

Original error was: libstdc++.so.6: cannot open shared object file: No such file or directory
```

justfile
jerbaroo: Reverted the timeout changes in this branch.

opsqueue/benches/chunks_select.rs
jerbaroo: I have updated the code to support extending with chunks. Though I didn't flip the order, so now it looks like:

```
/// For each shape:
///   For each amount of chunks:
///     Seed (all chunks) or extend (additional chunks) the database
///     FOR EACH strategy:
///         Create new dispatcher.
///         Run benchmark
///         Calculate stats
///         Write result
```

Notably we cannot do the extending for the `FewSubmissionsManyChunks` shape.

justfile
```diff
@@ -73,11 +68,23 @@ nix-test-integration *TEST_ARGS: nix-build
...
+# Benchmark query cost and render a graph.
+[group('bench')]
+bench-chunks-select:
+  #!/usr/bin/env bash
+  set -euo pipefail
+  cargo bench --bench chunks_select
+
+  cd libs/opsqueue_python
+  source "./.setup_local_venv.sh"
+  cd -
```
ReinierMaas: It's painfull that we can't `source ./libs/opsqueue_python/setup_local_venv.sh` directly without `cd`ing to the correct working directory...

opsqueue/benches/chunks_select.rs
```diff
@@ -0,0 +1,320 @@
...
+    )) = current_layout
+    {
+        // We can only extend if the shape of the data is compatible with the current inserted data.
+        if current_submissions_per_metadata_value == submissions_per_metadata_value
+            && current_chunks_per_submission == chunks_per_submission
+        {
+            we_are_extending = true;
+            current_metadata_values = current_metadata_values_;
+        }
+    }
```
ReinierMaas: This can be written with fewer mutations:

```suggestion
    let (we_are_extending, current_metadata_values) = if let Some((
        current_metadata_values,
        current_submissions_per_metadata_value,
        current_chunks_per_submission,
    )) = current_layout
    && current_submissions_per_metadata_value == submissions_per_metadata_value
    && current_chunks_per_submission == chunks_per_submission
    {
        (true, current_metadata_values)
    } else { (false, 0) }
```

opsqueue/benches/chunks_select.rs
```diff
@@ -0,0 +1,320 @@
...
+}
+
+/// Runs the selection query and fetches just the first chunk.
+async fn fetch_and_reserve_chunk(
+    db_pools: &db::DBPools,
+    strategy: Strategy,
+    dispatcher: &Dispatcher,
+) -> ChunkId {
+    use tokio::sync::mpsc::unbounded_channel;
+    let (notifier, _) = unbounded_channel();
```
ReinierMaas, line 210: Using an underscore here `_` drop the receiving side instantly. The sending side will receive errors when it tries to send anything over this channel. I haven't looked at the underlying code but does the reservation hold if we drop this receiving side of the channel? We could opt to return the receiving side together with the `ChunkId` and retain them until we finish our reservation to ensure that we can _receive_ messages from the view of the reserver.

opsqueue/benches/chunks_select.rs
```diff
@@ -0,0 +1,320 @@
...
+    let reserved = dispatcher
+        .fetch_and_reserve_chunks(db_pools.reader_pool(), strategy, 1, &notifier)
+        .await
+        .unwrap();
+    if let [(chunk, _)] = reserved.as_slice() {
+        ChunkId::from((chunk.submission_id, chunk.chunk_index))
```
ReinierMaas, line 216: What are we dropping here? Should we retain that until we finish our chunk reservation?

opsqueue/benches/chunks_select.rs
```diff
@@ -0,0 +1,320 @@
...
+async fn bench_strategy(
+    db_pools: &db::DBPools,
+    strategy: &Strategy,
+    dispatcher: &Dispatcher,
+) -> BenchStats {
+    let mut all_reservation_durations: Vec<Vec<f64>> = Vec::with_capacity(SAMPLES_PER_STAT);
+    // For each of the samples we want to collect.
+    for _ in 0..SAMPLES_PER_STAT {
+        let mut sample_reservation_durations = Vec::with_capacity(RESERVATIONS_IN_SAMPLE);
+        let mut chunks_reserved = Vec::new();
```
ReinierMaas:
```suggestion
        let mut chunks_reserved = Vec::with_capacity(RESERVATIONS_IN_WARMUP + RESERVATIONS_IN_SAMPLE);
```

You can also combine this with allocating the vector once and using `.drain(..)` when looping over the reserved chunks to retain the vectors allocation.

opsqueue/benches/chunks_select.rs
```diff
@@ -0,0 +1,320 @@
...
+        }
+    }
+    BenchStats::new(all_reservation_durations)
+}
+
+fn main() {
+    let runtime = tokio::runtime::Runtime::new().unwrap();
+    let csv_path = PathBuf::from("benches/chunks_select_bench.csv");
+    if let Some(parent) = csv_path.parent() {
+        let _ = std::fs::create_dir_all(parent);
```
ReinierMaas: You probably just want to `.unwrap()`` here. This should succeed if the parents exist and you want to early out if it fails to create the parent directories.

ReinierMaas (approved):
LGTM

I left a few smaller changes you could still make.

justfile
jerbaroo: Good point, no longer needed in  8a44797e2926a5256149cb8ef9d1ece6c310285b

opsqueue/benches/chunks_select.rs
jerbaroo: The reservation holds if we drop the receiving side of the channel.

The sending side of the channel is not used during the reservation except for being stored in the `Reserver` so that it can be used when the reservation expires.

jerbaroo: On reservation timeout if a sending error occurs it is just dropped:

```rs
                    // Only error case is if receiver is no longer listening
                    // In that case, nobody cares about the value being evicted anymore.
                    // So `let _ =` is correct here.
                    let _ = val.1.send(val.0);

```

opsqueue/benches/chunks_select.rs
jerbaroo: In `fetch_and_reserve_chunks` the stream of chunk that are successfully reserved are joined with submission info. `reserved` is a `Vec<(Chunk, Submission)>`, but here with `as_slice` the second value of the tuple is only a reference to the `Submission`, so nothing is dropped.

opsqueue/benches/chunks_select.rs
jerbaroo: Nice!, applied in 2e56142fc0fe76eeaf2927916d3d98ed1fb320d8

opsqueue/benches/chunks_select.rs
jerbaroo: Good idea, avoiding the re-allocation in cc114c61ad172d418ab470ef73e55634e3e2d9c8

opsqueue/benches/chunks_select.rs
jerbaroo: Early out in 0da77294c9d275e04d996833e382452c5ce4badc

jerbaroo:
@OpsBotPrime merge

OpsBotPrime:
Rebased as 7f83e4f22b35ba8ee11b55c59485e871873d9d78, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/c67f6923-b76a-413d-b9cb-407c6f572ce3?pipeline_id=f997e7b3-27e5-436e-9f82-2e485fb54c5c) started.

PR #152, opened 2026-07-24
SQLX: Keep database in sync with migrations

ReinierMaas:
I noticed that the checked-in database wasn't up-to-date with the latest migrations. This enforces the database to remain up-to-date.


default.nix
```diff
@@ -32,6 +32,7 @@ let
       pkgs.biome
       pkgs.nixfmt
       pkgs.pre-commit
+      pkgs.python3Packages.identify
```
jerbaroo: I don't see how this is needed, suggest add a comment

.pre-commit-config.yaml
```diff
@@ -44,6 +44,18 @@ repos:
         language: system
         types: [nix]
         exclude: extra
+      - id: sqlx
+        name: SQLX Database up-to-date
+        entry: >-
+          bash -euo pipefail -c 'DB_URL="sqlite://./opsqueue/opsqueue_example_database_schema.db";
+          if [[ -f ./opsqueue/opsqueue_example_database_schema.db ]]; then
+            sqlx migrate run --database-url "$DB_URL" --source ./migrations;
```
jerbaroo: suggest only use the condition to determine the sqlx command `migrate/database`. Avoids the reader needing to scan the two commands seeing how they are different

default.nix
ReinierMaas: It is necessary if I wanted to to use `types` instead of `files` but the version of `identify` from our `nixpkgs` pin isn't new enough to support `sqlite` (it does support `sql` but then doesn't trigger on `.db`...) I now simply use the file filter to make it simple to understand.

ReinierMaas:
@OpsBotPrime merge!

OpsBotPrime:
Rebased as c88ff62238cd912a70fc5d31fe8d918ae97955dd, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/0fee232d-329e-4902-8abb-3c0f40d58015?pipeline_id=78a9ea4b-c517-4654-b7fe-2919a569a00f) started.

PR #185, opened 2026-09-18
Add some consumer client docs

jerbaroo:
Wrote some comments as I was trying to understand this code. Hopefully it helps a future reader.


opsqueue/src/consumer/client.rs
```diff
@@ -361,7 +367,10 @@ impl Client {
                             break;
                         }
                         Some(Err(e)) => {
-                            tracing::error!(error = as_dyn_error(&e), "Opsqueue consumer client background task closing, reason");
+                            tracing::error!(
+                                error = as_dyn_error(&e),
+                                "Opsqueue consumer client background task closing: {e}"
+                            );
                             break;
```
ReinierMaas: Could you revert this change? We intentionally do not include the error in the message here.

What happens is that the error is reported to Sentry and the error chain is connected via the `as_dyn_error` which renders every error in the chain separately. By including it in the message the underlying errors are reported twice (and thrice for three levels deep etc.)

ReinierMaas (approved):
LGTM! Thanks for the additional documentation. I only require one small revert.

